PR #43, opened 2017-10-25
NAP-200 Finnish terms fixed

rckopteri:
If you agree that this can be merged. Go ahead and merge :)


ote/resources/public/language/fi.edn
```diff
@@ -149,7 +149,7 @@
 
   :ote.db.transport-service/information-service-accessibility
   {:audio-for-hearing-impaired "Ääniopastus kuulovammaisille"
-   :audio-information "Äänelliset informaatiot"
+   :audio-information "Kuulutukset"
```
K1ll3rF0x: This is not necessarily "kuulutukset". What about billboards that provide audio guidance when pressing a button etc. But, this is good enough for publishing.

PR #44, opened 2017-10-25
Nap 167 check out owners rights

rckopteri:
Hox, to make those unit tests work again, you need to update test data to the database.
ckan-group-id '79046442-ad25-4865-a174-ec199a4b39c4' is now required data for test operator.


ote/src/clj/ote/services/transport.clj
```diff
@@ -105,6 +106,9 @@
 (defn- save-transport-operator [db data]
   (upsert! db ::transport-operator/transport-operator data))
 
+(defn- user-has-rights-to-save-service [operator service]
```
K1ll3rF0x: Could we have more generic access control functions that can be reused in every form? ->  user-has-rights-to-service [operator service].
Getting ids from operator and services seems to be using quite specific paths. Could we have generic helper fns for those tasks? e.g get-id

rckopteri: You have to give me example how to do this.
I have no idea how to create more generic method for access rights.

I moved access right check to cljc folder. Now it is possible to take into account in ui side too (not implemented, but possible).

PR #49, opened 2017-10-30
NAP-210 service hours fix

rckopteri:


ote/src/cljc/ote/time.cljc
```diff
@@ -24,6 +25,24 @@
          cljs-time/to-default-time-zone
          (format/unparse (format/formatter "dd.MM.yyyy HH:mm"))))))
 
+#?(:cljs
+   (defn js-time-to-string [time]
+     (if  (nil? time)
+       " " ;: if nil - print empty string
```
tatut: not an empty string, contains a single space.

ote/src/cljc/ote/time.cljc
```diff
@@ -24,6 +25,24 @@
          cljs-time/to-default-time-zone
          (format/unparse (format/formatter "dd.MM.yyyy HH:mm"))))))
 
+#?(:cljs
+   (defn js-time-to-string [time]
```
tatut: Isn't this formatting? Should it be called `format-js-time` or something like that. Or have a function `from-js-time` that turns JS Date object to OTE Time record and call format on that.

ote/src/cljc/ote/time.cljc
```diff
@@ -24,6 +25,24 @@
...
+     (if  (nil? time)
+       " " ;: if nil - print empty string
+       (->> time
+            cljs-time/to-default-time-zone
+            (format/unparse (format/formatter "HH:mm:ss"))))))
+
+#?(:cljs
+   (defn js-time-from-db-time [db-time]
```
tatut: Perhaps `to-js-time` or something like that.

ote/src/cljs/ote/ui/form_fields.cljs
```diff
@@ -194,6 +194,18 @@
...
+(defmethod field :time-picker [{:keys [update! ok-label cancel-label default-time] :as opts} data]
+  (let [time-picker-time (if (= nil? data) default-time data)]
+  [ui/time-picker
+   {:format "24hr"
+    :cancel-label cancel-label
+    :ok-label ok-label
+    :minutes-step 5
+    :default-time (time/js-time-from-db-time time-picker-time)
+    :on-change (fn [event value]
+                 (update! (with-meta (time/parse-time (time/js-time-to-string value))
```
tatut: Does a time-picker need an incomplete value? User can't type in this component.

ote/src/cljc/ote/time.cljc
rckopteri: We already have time/parse-time that converts string to ->Time object.
So I only change function name to format-js-time.

ote/src/cljc/ote/time.cljc
rckopteri: Nice catch. Thanks.

PR #52, opened 2017-10-31
NAP-196 Drawing geometries

tatut:
Add leaflet-draw plugin to the map. User can draw area with polygons, lines or markers. The area can be renamed.


ote/resources/public/index.html
```diff
@@ -7,6 +7,9 @@
         <link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css"
               integrity="sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ=="
               crossorigin=""/>
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet.draw/0.4.12/leaflet.draw.css"
```
rckopteri: We don't need leaflet.css in every page. Is there a way to include this only when needed?

tatut: We could inject it when we mount but that's more hassle than it's worth now.

PR #56, opened 2017-11-03
NAP-221 Support multiple external interfaces

tatut:


database/src/main/resources/db/migration/V1_11__multiple_external_interfaces.sql
```diff
@@ -0,0 +1,14 @@
+-- Allow multiple external interfaces to be defined
+
+CREATE TABLE "external-interface-description" (
+  id SERIAL PRIMARY KEY,
+  "transport-service-id" INTEGER REFERENCES "transport-service" (id),
+  "external-interface" service_link,
+  "format" TEXT,
+  "ckan-resource-id" TEXT REFERENCES "resource" (id)
```
rckopteri: Could you explain why external interfaces have connection to ckan resource? They already have a link to transport-service (transport-service-id) ?

tatut: When publishing a service, each external interface is saved as a CKAN resource. We need the id because we may want to update it later.

PR #73, opened 2017-11-10
152671132 uniformed header

rckopteri:


ote/src/cljs/ote/app/state.cljs
```diff
@@ -29,3 +29,17 @@
...
                          :geojson nil}))
+
+
+(defn windowresize-handler
+  "When window size changes set width and height to app state"
+  [event]
+  (let [w (.-innerWidth js/window)
+        h (.-innerHeight js/window)]
+    (.log js/console "resize")
+    (do
```
tatut: unnecessary `do` and both width and height can be `assoc`ed with the same call.
`(swap! app assoc :width w :height h)`

ote/src/cljs/ote/views/main.cljs
```diff
@@ -42,6 +42,8 @@
                    :primary-text user-name
                    :selected-text-color (color :grey900)
                    :on-click #(e! (fp-controller/->ChangePage :front-page))} ]
+    [ui/menu-item {:primary-text "Näytä profiili"
```
tatut: Use translations in menu item texts.
Use an event for the click handler.

ote/src/cljs/ote/views/main.cljs
```diff
@@ -55,31 +57,82 @@
...
-   [:a.ote-nav
-    {:class (is-topnav-active :own-services (:page app))
-     :href "#"
-     :on-click #(e! (fp-controller/->ChangePage :own-services))}
-    (tr [:common-texts :navigation-own-service-list]) ]
-   [:div.user-menu {:class (is-user-menu-active app) }
-    (r/as-element (user-menu e! (get-in app [:user :name])))]])
+   [:div.container-fluid
+
+      (if (< 900 (get app :width))
```
tatut: I think it would be easier to read the condition as `(> (:width app) 900)` and also maybe the magic number `900` should be moved to a style ns and given a name (like `mobile-width-px`?)

ote/src/cljs/ote/views/main.cljs
```diff
@@ -116,4 +169,28 @@
...
        [:div.row
-        [debug/debug app]])]]]))
+        [debug/debug app]])
+
+     [:footer.site-footer
+      [:div.container-fluid
+       [:div.row
+        [:div.col-md-2.footer-links
+          [:a.logo {:href "#" }
+            [:img {:src "/livi_logo_valkoinen.svg" :alt "Liikennevirasto Logo"}]]]
```
tatut: Use translations for alt texts

ote/src/cljs/ote/app/state.cljs
rckopteri: Fixed

ote/src/cljs/ote/views/main.cljs
rckopteri: Translated, but I didn't add correct event yet. 
Profile link is different for different users. And I need to investigate it a bit more.

ote/src/cljs/ote/views/main.cljs
rckopteri: Changed

ote/src/cljs/ote/views/main.cljs
rckopteri: Translated

PR #80, opened 2017-11-14
152630496 Emailin käyttäminen ckan tunnuksen sijasta

K1ll3rF0x:
* Allow only unique email addresses. Effectively forces one email per account.
* Implement login with email or username.
* Implement password resetion with email or username.
* Update CKAN translation, and fix old translation errors.


nap/docker/ckan/ckan-plugins/ckanext-napote_theme/ckanext/napote_theme/templates/user/snippets/login_form.html
```diff
@@ -0,0 +1,31 @@
...
+#}
+{% import 'macros/form.html' as form %}
+
+{% set username_error = true if error_summary %}
+{% set password_error = true if error_summary %}
+
+<form action="{{ action }}" method="post" class="form-horizontal">
+    {{ form.errors(errors=error_summary) }}
+
+    {{ form.input('login', label=_("Username or email"), id='field-login', value="", error=username_error, classes=["control-medium"]) }}
```
rckopteri: Not a biggie, but: In the implementation logic you first check if user gave email and if email can't be found username is used. But here the help text indicates that username is used first.

K1ll3rF0x: The order of the check has no affect how user logs in. You can use username or email.

PR #82, opened 2017-11-14
Leaflet tries too soon to get bounds from layers. Add timeout to make…

rckopteri:
… it wait for 50ms before calculating bounds for zoom.

### DIFF
--- a/ote/src/cljs/ote/ui/leaflet.cljs
+++ b/ote/src/cljs/ote/ui/leaflet.cljs
@@ -62,14 +62,18 @@
     (when-let [bounds @bounds]
       (.fitBounds leaflet bounds))))
 
+(defn update-leaflet-bounds-soon [run-soon]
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/ui/leaflet.cljs
```diff
@@ -62,14 +62,18 @@
...
 
+(defn update-leaflet-bounds-soon [run-soon]
+  (.setTimeout js/window #(run-soon) 50)
+  )
+
 (defn update-bounds-from-layers [this]
   (let [^js/L.map
   leaflet (aget this "refs" "leaflet" "leafletElement")]
-    (update-map-bounds-from-layers leaflet)))
+    (update-leaflet-bounds-soon #(update-map-bounds-from-layers leaflet))))
```
tatut: The function to run is the same in both invocations, just give `leaflet` as parameter to update-leaflet-bounds-soon and move the code thete.

PR #85, opened 2017-11-14
[WIP] 152601149 form for rental services

mmarkuss:
Basic fields in rental service view. Rental specific fields missing.


ote/resources/public/language/fi.edn
```diff
@@ -95,7 +95,7 @@
 
 
   :transport-service
-  {:ote.db.transport-service/name "Palvelun nimi"
+  {:ote.db.transport-service/name "Palvelun mimi"
```
K1ll3rF0x: Fix "mimi"

ote/src/cljs/ote/views/rental.cljs
```diff
@@ -9,39 +9,42 @@
             [ote.ui.buttons :as buttons]
             [ote.app.controller.transport-service :as ts]
             [ote.app.controller.rental :as rental]
-            [ote.db.transport-service :as transport-service]
+            [ote.db.transport-service :as t-service]
```
K1ll3rF0x: General comment for all developers. Maybe we should name X db, controller and view related imports better. When reading the code, it is more difficult to see what is transport-service db reference and what is controller reference.

tatut: This needs more discussion so that this important point isn't buried in a single PR comment. I fully agree, and currently we've been using `t-service` and `t-operator` aliases for the main db namespaces pretty consistently.

PR #95, opened 2017-11-17
CKAN user flow changes.

rckopteri:
Login -> Front page
Register -> Front Page
Dashboard -> Dashboard dataset
Removed group tab from Dashboard
Dashboard activity feed moved to be the last tab on that page


nap/docker/ckan/ckan-plugins/ckanext-napote_theme/ckanext/napote_theme/controller.py
```diff
@@ -33,6 +35,14 @@
 unflatten = dictization_functions.unflatten
 
 
+def set_repoze_user(user_id):
```
tatut: import this from other module

rckopteri: imported

PR #99, opened 2017-11-20
152884765 Fix service delete and add confirmation

tatut:
When deleting a published service, call CKAN API to do the delete (via "dataset_purge" API call) which cascades to all OTE resources.

When deleting an unpublished, delete from OTE database directly.

### DIFF
--- a/ote/src/cljs/ote/views/front_page.cljs
+++ b/ote/src/cljs/ote/views/front_page.cljs
@@ -30,15 +30,17 @@
       {:open true
        :title (tr [:dialog :delete-transport-service :title])
        :actions [(r/as-element
-                  [ui/raised-button
+                  [ui/flat-button
                    {:label (tr [:buttons :cancel])
                     :primary true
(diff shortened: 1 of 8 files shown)


ote/src/cljs/ote/views/front_page.cljs
```diff
@@ -16,8 +16,30 @@
...
+     [ui/dialog
+      {:open true
+       :title (tr [:dialog :delete-transport-service :title])
+       :actions [(r/as-element
+                  [ui/flat-button
+                   {:label (tr [:buttons :cancel])
+                    :primary true
+                    :on-click #(e! (ts/->CancelDeleteTransportService id))}])
+                 (r/as-element
+                  [ui/flat-button {:label (tr [:buttons :delete])
```
rckopteri: Regardless of the original material ui example I would suggest that Save/Delete/Ok buttons should be raised-buttons in modals.

PR #102, opened 2017-11-20
152924732 header tuning

K1ll3rF0x:
Hide "Omat palvelutiedot" header menu item from logged out users in CKAN and OTE.

### DIFF
--- a/ote/src/cljs/ote/views/main.cljs
+++ b/ote/src/cljs/ote/views/main.cljs
@@ -71,21 +71,20 @@
 
 (defn header-links [app]
   (filter some?
-          (conj
-            [{:page  :front-page
-              :label [:common-texts :navigation-front-page]
-              :url   "/"}
+          [{:page  :front-page
+            :label [:common-texts :navigation-front-page]
+            :url   "/"}
 
-             {:page  :services
-              :label [:common-texts :navigation-dataset]}
+           {:page  :services
+            :label [:common-texts :navigation-dataset]}
 
-             {:page  :organizations
-              :label [:common-texts :navigation-organizations]
-              :url   "/organization"}]
+           {:page  :organizations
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/views/main.cljs
```diff
@@ -66,20 +69,23 @@
                 :style style-base/flash-message
                 :auto-hide-duration 5000}])
 
-(def header-links
-  [{:page :front-page
-    :label [:common-texts :navigation-front-page]
-    :url "/"}
+(defn header-links [app]
+  (filter some?
+          (conj
```
tatut: You don't need the `conj`. You can just `(filter some? [first-item second-item (when condition? third-item)])`

ote/src/cljs/ote/views/main.cljs
```diff
@@ -93,8 +99,9 @@
                {:href     "#"
                 :on-click #(e! (fp-controller/->GoToUrl "/"))})
         [:img {:src "img/icons/nap-logo.svg" :style style-topnav/img }]]])
+
     (doall
-     (for [{:keys [page label url]} header-links]
+     (for [{:keys [page label url]} (header-links app)]
```
tatut: Alternatively to the filter done in header links, you could do:
```
(for [{:keys [page label url]} (header-links app)
      :when page]
 ...body...
```

and take out the filtering

ote/src/cljs/ote/views/main.cljs
K1ll3rF0x: True. Forgot to remove it after adding filter.

ote/src/cljs/ote/views/main.cljs
K1ll3rF0x: Better to keep collections clean from nil values.

PR #129, opened 2017-12-04
153130316 Fix IE Edge scroll issue

tatut:
Fix issue in localized text `rows` attribute that caused Edge to scroll.
Refactored the routes that lead to transport service edit forms.


tatut:
Please *try* this branch version before merging

PR #134, opened 2017-12-07
153208668 Table usability improvements

tatut:
Support new `:prepare-for-save` key for a form schemas that allows for processing before sending values to the server. This is useful because most tables want to add post processing that removes empty values.

**Move effectively-empty? to new util ns**

**Fix values ns require**

**Force table to have one row, don't allow adding if last row is empty**

**Ignore some keys, add without-empty-rows helper**

**Fix naming of info fields**

**Add schemas to footer for use when preparing form data for save**

**Add schemas to footer params**

### DIFF
--- a/ote/src/cljs/ote/ui/form.cljs
+++ b/ote/src/cljs/ote/ui/form.cljs
@@ -178,7 +178,6 @@
   (reduce
    (fn [prepared-data {:keys [name read write prepare-for-save] :as s}]
      (let [value (prepare-for-save ((or read name) data))]
-       (.log js/console "PREPARE " (pr-str name) " = " (pr-str value))
        (if write
          (write prepared-data value)
          (assoc prepared-data name value))))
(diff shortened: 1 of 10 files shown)


ote/src/cljs/ote/ui/form.cljs
```diff
@@ -174,6 +174,16 @@
     ::first-modification (or first (t/now))
     ::latest-modification (t/now)))
 
+(defn prepare-for-save [schemas data]
+  (reduce
+   (fn [prepared-data {:keys [name read write prepare-for-save] :as s}]
+     (let [value (prepare-for-save ((or read name) data))]
+       (.log js/console "PREPARE " (pr-str name) " = " (pr-str value))
```
rckopteri: No logging needed here

ote/src/cljs/ote/views/passenger_transportation.cljs
```diff
@@ -210,7 +209,8 @@
                ::t-service/booking-service)
               (accessibility-group)
               (pricing-group e! form-data)
```
rckopteri: Should pricing-group get first empty row by default as well?

PR #135, opened 2017-12-07
153130249 edge remove service bug

rckopteri:
There is three fixes in this pr.

1. When spinner is shown whole page is rendered (footer and header) not empty page with spinner
2. :front-page page is now removed and we cannot redirect users to :front-page anymore
3. When user deletes a service, we now set :service-changed? flag to true, so we can refresh app-state completely when needed.

### DIFF
--- a/ote/src/cljs/ote/app/controller/front_page.cljs
+++ b/ote/src/cljs/ote/app/controller/front_page.cljs
@@ -38,9 +38,9 @@
        (comm/post! "transport-operator/data" {}
                   {:on-success (tuck/send-async! ->TransportOperatorDataResponse)
                    :on-failure (tuck/send-async! ->TransportOperatorDataFailed)})
-      (assoc app :transport-operator-data-loaded? false
-                 :services-changed? false)
-       ) app))
+       (assoc app :transport-operator-data-loaded? false
+                  :services-changed? false))
+     app))
 
 (extend-protocol tuck/Event
 
@@ -77,7 +77,7 @@
 
   EnsureTransportOperator
   (process-event [_ app]
-     (if (get app :services-changed?)
+     (if (:services-changed? app)
       (get-transport-operator-data app)
       app))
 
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/app/controller/front_page.cljs
```diff
@@ -32,6 +32,16 @@
...
 
+(defn get-transport-operator-data [app]
+   (if (get app :transport-operator-data-loaded? true)
+     (do
+       (comm/post! "transport-operator/data" {}
+                  {:on-success (tuck/send-async! ->TransportOperatorDataResponse)
+                   :on-failure (tuck/send-async! ->TransportOperatorDataFailed)})
+      (assoc app :transport-operator-data-loaded? false
+                 :services-changed? false)
+       ) app))
```
tatut: confusing paren placement, put closing paren on previous line, so that `app` as the else clause is clear

ote/src/cljs/ote/app/controller/front_page.cljs
```diff
@@ -67,11 +77,8 @@
 
   EnsureTransportOperator
   (process-event [_ app]
-    (if (empty? (get app :transport-operator))
-      (do
-        (routes/navigate! :no-operator)
-        (assoc app :page :no-operator)
-        )
+     (if (get app :services-changed?)
```
tatut: It is clearer to use `(:services-changed? app)` instead of using `get` function. Or just destructure the wanted key in the parameter declaration,

ote/src/cljs/ote/app/controller/transport_service.cljs
```diff
@@ -284,8 +284,9 @@
   DeleteTransportServiceResponse
   (process-event [{response :response} app]
     (let [filtered-map (filter #(not= (:ote.db.transport-service/id %) (int response)) (get app :transport-service-vector))]
```
tatut: do we still need to filter this collection if we are fetching the data again?

rckopteri: Nope, we don't need.
But delete isn't working correctly, so I need to ensure functionality first.

PR #158, opened 2017-12-13
153474070 CKAN/OTE flow changes

tatut:
Change the CKAN dashboard (remove my services as OTE has a better view for it). Create organizations in OTE (via CKAN API). This removes the need to have 2 different forms for creating an org.

**Hide pictures from org list**

**Buttons for adding new transport operator**

**Create CKAN org**

**Event for creating a transport operator**

**Add helper for two column table**

**Support special :divider option in select list**

**Force postal code to be valid with regex**

**Remove org edit from user menu**

**Refactor front page operator selection**

**Add authorization check editing transport operator**

**Remove image from transport operator**

**Set services changed flag immediately**
When returning from creating without saving, this makes sure no stale
operator data is in app state

**Remove add organization link**

### DIFF
--- a/ote/src/cljs/ote/views/front_page.cljs
+++ b/ote/src/cljs/ote/views/front_page.cljs
@@ -20,7 +20,8 @@
             [ote.style.base :as style-base]
             [reagent.core :as r]
             [ote.ui.form-fields :as form-fields]
-            [ote.ui.common :as ui-common]))
+            [ote.ui.common :as ui-common]
+            [ote.views.transport-operator :as t-operator-view]))
 
 (defn- delete-service-action [e! {::t-service/keys [id name]
                                   :keys [show-delete-modal?]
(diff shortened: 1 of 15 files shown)


ote/src/cljs/ote/views/front_page.cljs
```diff
@@ -93,39 +94,53 @@
...
+                        (::t-operator/name %))
+        :update!   #(if (nil? %)
+                      (e! (to/->CreateTransportOperator))
+                      (e! (to/->SelectOperator %)))
+        :options     (into (mapv :transport-operator operators)
+                           [:divider nil])
+        :auto-width? true}
+       operator]
+      [ui/flat-button {:label (tr [:buttons :edit])
+                       :style {:margin-top "1.5em"
```
rckopteri: could use stylefy

PR #159, opened 2017-12-13
153561656 organization ui tuning

rckopteri:


ote/resources/public/language/sv.edn
```diff
@@ -424,6 +424,11 @@
  :organization-page
  {:contact-types            "Yhteystavat"
   :organization-form-title  "Uppassningenstuottajan tiedot"
+  :help-desc-1              "Tiedot tulevat näkyviin NAP-palvelun julkisille sivuille."
```
tatut: You could just use single :help-desc that is a vector of parts and parameterized.

PR #167, opened 2017-12-15
153171568 ie save service problem

rckopteri:
IE 11 is working now.
But, if you have time, you can test this on your own machine.


ote/src/cljs/ote/ui/form_fields.cljs
```diff
@@ -259,7 +263,7 @@
      (doall
       (map-indexed
        (fn [i option]
-         (let [checked? (selected option)]
+         (let [checked? (if (nil? (selected option)) false true)]
```
tatut: you can just coerce to boolean: `(boolean (selected option))`

rckopteri: changed

PR #170, opened 2017-12-15
153698015 add common loading indicator

K1ll3rF0x:


ote/src/cljs/ote/communication.cljs
```diff
@@ -14,32 +17,47 @@
...
+
+  (progress/start)
+  (swap! query-counter inc)
   (POST (request-url url)
-        {:params body
-         :handler on-success
-         :error-handler on-failure
-         :format (transit-request-format)
+        {:params          body
+         :handler         (comp handle-response! on-success)
```
tatut: Minor issue.
While `comp` works, `handle-response!` doesn't take any parameters. This implies that handle-response takes the return value of on-success/on-failure.

I think it would be clearer if handle-response takes the success/failure as parameter and returns a handler which calls the success/failure as the last step.

PR #175, opened 2017-12-18
Update own service page 

rckopteri:
Change add new service button text
Add Add new service button to the bottom of own service page when there is no own services
Add red info box to inform users that this is test server.


ote/src/cljs/ote/views/front_page.cljs
```diff
@@ -101,13 +101,19 @@
...
                         :on-click #(do
                                      (.preventDefault %)
                                      (e! (ts/->OpenTransportServiceTypePage)))
                         :primary  true
-                        :icon (ic/content-add)}]]]
+                        :icon (ic/content-add)}]]
+     [:div {:style {:border "red 4px dashed"}}
+      [:p {:style {:padding "10px"}} "HUOM: Tämä on NAP -palvelun testiversio."
+       [:br]
+       "Tiedot eivät siirry varsinaiseen NAP-palveluun, joka avataan tuotannossa 20.12.2017."]]]
```
tatut: This needs to made conditional as it must not be shown in the production environment.

K1ll3rF0x: We could pass dev (or is staging currectly) / production param, and use that in the conditional.

PR #203, opened 2017-12-21
153843642 sub type problems

rckopteri:


ote/src/clj/ote/services/transport.clj
```diff
@@ -236,7 +236,7 @@
   [nap-config db user {places ::t-service/operation-area
                        external-interfaces ::t-service/external-interfaces
                        :as data}]
-  ;(println "DATA: " (pr-str data))
+  (println "DATA: " (pr-str data))
```
K1ll3rF0x: Comment out?

PR #204, opened 2017-12-21
153598636 Add custom leaflet zoom controls and translate zoom buttons

K1ll3rF0x:


ote/src/cljs/ote/views/place_search.cljs
```diff
@@ -104,6 +107,7 @@
         [leaflet/Map {;;:prefer-canvas true
                       :ref "leaflet"
                       :center #js [65 25]
+                      :zoomControl false
```
rckopteri: I assume, that this doesn't  prevent zooming but hides default zoom buttons?

K1ll3rF0x: Yes, it allows us to use customized zoom controller.

PR #211, opened 2018-01-02
153737273 terminal access change

rckopteri:
Add place description and require notification checkbox to terminal assistance data


ote/src/cljs/ote/views/terminal.cljs
```diff
@@ -28,13 +28,12 @@
    (tr [:field-labels :terminal ::t-service/indoor-map])
    ::t-service/indoor-map))
 
-(defn- assistance-service-group []
+(defn- assistance-service-group [data]
```
tatut: remove unused parameter here and call site

PR #213, opened 2018-01-02
153989516 ie11 cache response problem

rckopteri:


ote/src/clj/ote/components/http.clj
```diff
@@ -115,6 +115,14 @@
    :headers {"Content-Type" "application/json+transit"}
    :body (transit/clj->transit data)})
 
+(defn no-cache-transit-response
+  "Return the given Clojure `data` as a Transit response with status code 200."
+  [data]
+  {:status 200
+   :headers {"Content-Type" "application/json+transit"
```
tatut: There's minor duplication here, perhaps `transit-response` should take an optional map of extra headers to set, which `no-cache-transit-response` could call with a map that would be merged.

ote/src/clj/ote/components/http.clj
```diff
@@ -115,6 +115,14 @@
    :headers {"Content-Type" "application/json+transit"}
    :body (transit/clj->transit data)})
 
+(defn no-cache-transit-response
+  "Return the given Clojure `data` as a Transit response with status code 200."
+  [data]
+  (let [response (transit-response data)
+        headers (merge (get response :headers)  {"Cache-Control" "no-cache, no-store"})
+        response (assoc response :headers headers)]
+    response))
```
tatut: `(update (transit-response data) :headers merge {"Cache-Control" "no-cache, no-store"})`

PR #222, opened 2018-01-05
Add 24h-allday checkbox value to db

rckopteri:


ote/src/cljs/ote/views/parking.cljs
```diff
@@ -86,7 +86,7 @@
 
 (defn service-hours-group [e!]
   (let [tr* (tr-key [:field-labels :service-exception])
-        write (fn [key]
+        write-time (fn [key]
                 (fn [{all-day? ::t-service/all-day :as data} time]
```
K1ll3rF0x: all-day? is also used in rental form.

ote/src/cljs/ote/views/parking.cljs
```diff
@@ -113,31 +113,26 @@
                    {:name  ::t-service/all-day
                     :width "10%"
                     :type  :checkbox
-                    :write (fn [data all-day?]
+                    :write (do
+                             #(assoc-in %1 [::t-service/service-hours ::t-service/all-day] %2)(fn [data all-day?]
```
K1ll3rF0x: Why is assoc-in here?

ote/src/cljs/ote/views/transport_service_common.cljs
```diff
@@ -251,31 +251,27 @@
        {:name ::t-service/all-day
         :width "10%"
         :type :checkbox
-        :write (fn [data all-day?]
+        :write (do
+                 #(assoc-in %1 [::t-service/service-hours ::t-service/all-day] %2)
```
K1ll3rF0x: Why assoc-in?

ote/src/cljs/ote/views/parking.cljs
rckopteri: Yes. Rental uses the same common component that passenger-transportation uses

ote/src/cljs/ote/views/parking.cljs
rckopteri: yes. it didn't affect I thought that if :write parameter is given it will override default functionality. Thanx.
Removed now

PR #223, opened 2018-01-05
Delete published service from OTE even if service cannot be found fro…

mmarkuss:
…m CKAN.


K1ll3rF0x:
https://www.pivotaltracker.com/story/show/154049350

PR #231, opened 2018-01-10
153813187 external interfaces cannot be deleted

mmarkuss:


ote/src/clj/ote/services/transport.clj
```diff
@@ -266,6 +272,7 @@
 
     ;; If published, use CKAN API to add dataset and resource
     (when (::t-service/published? data)
+      (publish/delete-resources-from-published-service! nap-config user (filter #(:deleted? %) external-interfaces))
```
rckopteri: This row would be easier to read if that filtered result were in if. 
like this:
(when (filter #(:deleted? %) external-interfaces))
  publish/delete-resources-from-published-service! nap-config user (filter #(:deleted? %) external-interfaces)))

PR #237, opened 2018-01-12
154237443 Use unique JS name based on git revision SHA

tatut:
**Buildtime tools to get git revision and rename JS file**

**Determina JS location by buildtime git SHA**

**Add postbuild action**

### DIFF
--- a/ote/src/clj/ote/tools/git.clj
+++ b/ote/src/clj/ote/tools/git.clj
@@ -1,0 +1,7 @@
+(ns ote.tools.git
+  "Tools for Git in build process"
+  (:require [clojure.java.shell :as sh]
+            [clojure.string :as str]))
+
+(defmacro current-revision-sha []
+  {:current-revision-sha (str/trim (:out (sh/sh "git" "rev-parse" "HEAD")))})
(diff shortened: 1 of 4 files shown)


ote/src/clj/ote/tools/git.clj
```diff
@@ -0,0 +1,7 @@
+(ns ote.tools.git
+  "Tools for Git in build process"
+  (:require [clojure.java.shell :as sh]
+            [clojure.string :as str]))
+
+(defmacro current-revision-sha []
```
rckopteri: We don't win anything by using git revisions but add complexity.
I know it won't hurt us but still take into consideration that we can use something simpler, e.g. create hash from timestamp.

tatut: git revision uniquely identifies the current version and is imo much better than a timestamp. It is also simpler because the git revision does not change during the build (like a timestamp does)

K1ll3rF0x: webpack style hashing is also viable: https://github.com/alexindigo/webpack-chunk-hash

PR #240, opened 2018-01-13
154305299 Login to OTE

tatut:
Implemented a login form in the OTE application and a backend service that can use CKAN database to log in.


ote/src/cljs/ote/app/controller/login.cljs
```diff
@@ -0,0 +1,90 @@
+(ns ote.app.controller.login
```
rckopteri: Next step is to create/register users as well. So consider renaming this controller to user. And we can add user creation functions to the same controller.

PR #248, opened 2018-01-18
154205786 Logging improvements, take 2

tatut:
**Configure logging level**
Add slf4j timbre so that Java libraries that use slf4j will log
through timbre and respect the same logging configuration.

**Set stg/prod log level to INFO**

### DIFF
--- a/aws/ansible/templates/ote/config.edn.j2
+++ b/aws/ansible/templates/ote/config.edn.j2
@@ -11,4 +11,6 @@
            :name-field "NAMEFIN"
            :url "https://s3.eu-central-1.amazonaws.com/ote-assets/places/finnish-municipalities/SuomenKuntajako_2017_10k.shp"}}
  :nap {:api "{{base_url}}/api/"
-       :export-base-url "{{base_url}}"}}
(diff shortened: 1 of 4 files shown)


aws/ansible/templates/ote/config.edn.j2
```diff
@@ -11,4 +11,6 @@
            :name-field "NAMEFIN"
            :url "https://s3.eu-central-1.amazonaws.com/ote-assets/places/finnish-municipalities/SuomenKuntajako_2017_10k.shp"}}
  :nap {:api "{{base_url}}/api/"
-       :export-base-url "{{base_url}}"}}
+       :export-base-url "{{base_url}}"}
+ :log {:level :info}
```
rckopteri: could we use warning level in production?

tatut: we could but should we?

rckopteri: I think that we only need production logs when we have problems and warnings and errors are enough in those situations.
So I would suggest that we log only things we need and not everything

PR #255, opened 2018-01-18
153863629 Document title and favicon

tatut:
**Add favicon resources**

**Set document title based on page**


ote/src/cljs/ote/views/main.cljs
```diff
@@ -317,6 +317,20 @@
      [login-form e! login]]
     [login-action-cards]]])
 
+(defn document-title [page]
+  (set! (.-title js/document)
```
rckopteri: Why not add title to the <head> element directly?

PR #265, opened 2018-01-23
154389818 admin can remove published services

rckopteri:
# Added
 - Added possibility for admin users to remove published services.
 - Added auditlog where all actions that admins do will be stored.

# Environment
 - Added new table auditlog, new composite type auditlog_event_attribute and new enum auditlog_event_type


ote/src/clj/ote/services/admin.clj
```diff
@@ -20,15 +24,33 @@
...
                             :name (str "%" query "%")}))
 
-(defn- admin-routes [db http]
+(defn- admin-delete-transport-service!
+  "Allow admin delete single transport service by id"
+  [nap-config db user id]
+  (let [return (transport/delete-transport-service! nap-config db user id)
+        auditlog {::auditlog/event-type :delete-service
+                  ::auditlog/event-attributes
+                                      [{::auditlog/name "transport-service-id", ::auditlog/value (str id)}]
```
tatut: Add transport-service-name parameter that is the name of the service. That way we can later show a better description of what was deleted.

ote/test/clj/ote/services/transport_test.clj
```diff
@@ -175,18 +175,17 @@
...
+        terminal-service (http-get "admin" (str "transport-service/" (::t-service/id service)))
+        fetched (:transit terminal-service)
+        ;; Change operator id from 1 -> 2 - which will cause error
+        modified-terminal-service (assoc fetched ::t-service/transport-operator-id 2)]
 
     (is (thrown-with-msg?
-         clojure.lang.ExceptionInfo #"status 403"
-         (http-post "admin" "transport-service" modified-operator)))))
+          ;; Gives error, because admin has access to all operators, but there isn't operator with id 2
+          clojure.lang.ExceptionInfo #"status 500"
```
tatut: Maybe add a call with another user to also test the original 403 response still works.

PR #269, opened 2018-01-24
153170774 palvelun käytön analysointi

K1ll3rF0x:
# Added
* Google Analytics trackers for OTE and CKAN indexes.

### DIFF
--- a/ote/config.edn
+++ b/ote/config.edn
@@ -14,4 +14,5 @@
  :nap {:api "http://localhost:8080/api/"
        :export-base-url "http://localhost:8080"}
 
- :log {:level :debug}}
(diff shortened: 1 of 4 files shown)


ote/config.edn
```diff
@@ -14,4 +14,5 @@
  :nap {:api "http://localhost:8080/api/"
        :export-base-url "http://localhost:8080"}
 
- :log {:level :debug}}
+ :log {:level :debug}
+ :ga {:tracking-code "UA-113014221-1"}}
```
rckopteri: I thought that we agreed not to use google, but https://matomo.org/ :)

PR #270, opened 2018-01-25
 154625968 Enable feature flags

rckopteri:
# Environment
 - Add ote-login flag to config.edn
 - Add flag controller that can be used to check if feature is shown in production
 - Add check to login feature whether or not to show it in production

### DIFF
--- a/ote/config.edn
+++ b/ote/config.edn
@@ -16,5 +16,5 @@
 
  :log {:level :debug}
  :ga {:tracking-code "UA-113014221-1"}
- :enabled-features "example-feature, " ;; add features to the string, split by "," to enable them in production
+ :enabled-features #{:example-feature} ;; add features to the set to enable them in production
  }
--- a/ote/config.edn
+++ b/ote/config.edn
@@ -15,4 +15,6 @@
        :export-base-url "http://localhost:8080"}
 
  :log {:level :debug}
- :ga {:tracking-code "UA-113014221-1"}}
+ :ga {:tracking-code "UA-113014221-1"}
+ :flags {:new-login false} ;; set true if new feature is shown in production
+ }
--- a/ote/config.edn
+++ b/ote/config.edn
@@ -16,5 +16,5 @@
 
  :log {:level :debug}
  :ga {:tracking-code "UA-113014221-1"}
- :flags {:new-login false} ;; set true if new feature is shown in production
+ :enabled-features "example-feature, " ;; add features to the string, split by "," to enable them in production
  }
--- a/ote/config.edn
+++ b/ote/config.edn
@@ -16,5 +16,7 @@
 
  :log {:level :debug}
  :ga {:tracking-code "UA-113014221-1"}
- :enabled-features #{:example-feature} ;; add features to the set to enable them in production
+
+  ;; add features to the set to enable them in production
+ :enabled-features #{:ote-login}
  }
--- a/ote/src/clj/ote/services/index.clj
+++ b/ote/src/clj/ote/services/index.clj
@@ -37,7 +37,7 @@
 (defn index-page [config]
   (let [dev-mode? (:dev-mode? config)
         ga-conf (:ga config)
-        flags (:flags config)
+        flags (:enabled-features config)
         ]
     [:html
      [:head
@@ -72,7 +72,7 @@
 
       [:body {:id           "main-body"
               :onload       "ote.main.main();"
-              :new-login    (str (:new-login flags))
+              :features     flags
               :data-language localization/*language*}
       [:div#oteapp]
       (when dev-mode?
--- a/ote/src/clj/ote/services/index.clj
+++ b/ote/src/clj/ote/services/index.clj
@@ -37,8 +37,7 @@
 (defn index-page [config]
   (let [dev-mode? (:dev-mode? config)
         ga-conf (:ga config)
-        flags (str/join "," (:enabled-features config))
-        ]
+        flags (str/join "," (map name (:enabled-features config)))]
     [:html
      [:head
 
--- a/ote/src/clj/ote/services/index.clj
+++ b/ote/src/clj/ote/services/index.clj
@@ -37,7 +37,7 @@
 (defn index-page [config]
   (let [dev-mode? (:dev-mode? config)
         ga-conf (:ga config)
-        flags (:enabled-features config)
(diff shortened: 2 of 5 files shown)


ote/config.edn
```diff
@@ -15,4 +15,6 @@
        :export-base-url "http://localhost:8080"}
 
  :log {:level :debug}
- :ga {:tracking-code "UA-113014221-1"}}
+ :ga {:tracking-code "UA-113014221-1"}
+ :enabled-features "example-feature, " ;; add features to the string, split by "," to enable them in production
```
tatut: this would be better as a set `#{:example-feature :other-feature ...}`

ote/src/clj/ote/services/index.clj
```diff
@@ -34,9 +34,11 @@
    {:rel "mask-icon" :href "/ote/favicon/safari-pinned-tab.svg?v=E6jNQXq6yK" :color "#5bbad5"}
    {:rel "shortcut icon" :href "/ote/favicon/favicon.ico?v=E6jNQXq6yK"}])
 
-(defn index-page [dev-mode?]
-  (let [config (read-string (slurp "config.edn"))
-        ga-conf (:ga config)]
+(defn index-page [config]
+  (let [dev-mode? (:dev-mode? config)
+        ga-conf (:ga config)
+        flags (:enabled-features config)
```
tatut: and if the flags are a set, just `(str/join "," (:enabled-features config))`

ote/src/cljs/ote/app/controller/flags.cljs
```diff
@@ -0,0 +1,31 @@
...
+(defn is-in-production []
+  (let [page-url (-> (.-location js/window))]
+    (if (or (s/includes? (str page-url) "testi") (s/includes? (str page-url) "localhost"))
+      false
+      true)))
+
+(defn use-in-production?
+  "Check if given feature (f) is enabled in production"
+  [f]
+  (let [features  (str/split (-> js/document
```
tatut: `(into #{} (map keyword (str/split (.getAttribute js/document.body "features") #",")))`

tatut: If you parse it into a set, the `in?` function becomes unnecessary, just use the set as a function

ote/src/cljs/ote/views/main.cljs
```diff
@@ -165,13 +166,19 @@
                   (if desktop? style-topnav/desktop-link style-topnav/link))
                 {:style {:float "right"}})]]
        [:li
+        (if (flag-controller/use-in-production? "new-login")
```
tatut: use `:new-login´

ote/config.edn
rckopteri: I tried it. It is a string in that html page and I could not figure out how "#{:my-key}" is converted to set

ote/src/clj/ote/services/index.clj
rckopteri: ah, thanx

ote/src/clj/ote/services/index.clj
```diff
@@ -34,9 +34,11 @@
...
    {:rel "shortcut icon" :href "/ote/favicon/favicon.ico?v=E6jNQXq6yK"}])
 
-(defn index-page [dev-mode?]
-  (let [config (read-string (slurp "config.edn"))
-        ga-conf (:ga config)]
+(defn index-page [config]
+  (let [dev-mode? (:dev-mode? config)
+        ga-conf (:ga config)
+        flags (str/join "," (:enabled-features config))
+        ]
```
K1ll3rF0x: Fix square bracket position

ote/src/cljs/ote/app/controller/flags.cljs
```diff
@@ -0,0 +1,24 @@
+(ns ote.app.controller.flags
+  (:require [tuck.core :as tuck]
+            [ote.db.transport-service :as t-service]
+            [ote.localization :refer [tr tr-key]]
+            [clojure.string :as str]
+            [ote.communication :as comm]
+            [clojure.string :as s]))
+
+(defn true-string? [s] (= s "true"))
```
K1ll3rF0x: Remove. This is not needed.

ote/src/cljs/ote/views/main.cljs
```diff
@@ -165,13 +166,19 @@
                   (if desktop? style-topnav/desktop-link style-topnav/link))
                 {:style {:float "right"}})]]
        [:li
+        (if (flag-controller/use-in-production? ":new-login")
         [linkify "#" (tr [:common-texts :navigation-login])
```
K1ll3rF0x: Fix indent.

ote/src/clj/ote/services/index.clj
```diff
@@ -68,8 +70,10 @@
             window['ga-disable-" (:tracking-code ga-conf) "'] = true;
           }")))
 
-      [:body {:onload        "ote.main.main();"
-             :data-language localization/*language*}
+      [:body {:id           "main-body"
+              :onload       "ote.main.main();"
+              :features     flags
```
K1ll3rF0x: I find it strange that the feature flags are injected into html attribute as a string, whereas you could inject them as default values in the app state tree. 
If the flags would be in the state, we could also easily toggle them on/off in the browser console using some utility function, and test them out in live situation. This makes also sure that the flagged features are properly rendered if the state changes.

rckopteri: That would work too if we want to add feature flags to code. 

Currently we can't change app state without building new .jar file to the server. As long as feature flags are in config file we can change config and restart server. No new build is required. 

So we don't gain anything by changing the current implementation. It would only be different. And those values are just as easily changeable to the dom as app state is. JS code is required. In both situations.

Maybe the best option would be database based configs.

K1ll3rF0x: This works, so it is ok. But, this feels like a "karvalakki" approach. Passing flags through state or react props would be cleaner, but that is not critical. I'll stop my ranting now :)

tatut: Features often have implications on both the server and the client side, so they need to be passed in the initial render somehow. Attribute is a good enough solution that is easy to read in the client side without needing to make an AJAX call the get the flags config.

PR #273, opened 2018-01-26
Save and load service hours info for rentals.

mmarkuss:
# Fixed
- Rental pick-up location service hours info

# Datamodel
- Service hours info moved from rentals type to pick-up location type.

### DIFF
--- a/database/src/main/resources/db/migration/V1_49__rental_service_hours_info.sql
+++ b/database/src/main/resources/db/migration/V1_49__rental_service_hours_info.sql
@@ -1,0 +1,5 @@
+ALTER TYPE rental_provider_information
+ DROP ATTRIBUTE "service-hours-info";
+
+ALTER TYPE pick_up_location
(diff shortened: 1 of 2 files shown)


database/src/main/resources/db/migration/V1_49__rental_service_hours_info.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TYPE rental_provider_information
+ DROP ATTRIBUTE "service-hours-info";
```
rckopteri: Is someone added data to this column in production? Should we move datas to new column?

K1ll3rF0x: Good catch, we should check it.

mmarkuss: There is only one rental service including pick-up-location info. This service does not have service-hours-info. Therefore this migration should be safe.

PR #277, opened 2018-01-30
External interfaces field label changes and description data migration

K1ll3rF0x:
# Added
* A new "Data content" (Tietosisältö) multiselect field for external interfaces.

# Fixed
* Changes to external interfaces form fields and their labels.
* Migrated external interfaces license_url field data into description field.
* Removed licence_url field and its usages.


database/src/main/resources/db/migration/V1_50__external_interfaces_description.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TYPE interface_data_content AS ENUM (
```
rckopteri: number 50 migration is already in master. So rename migration file.

database/src/main/resources/db/migration/V1_50__external_interfaces_description.sql
```diff
@@ -0,0 +1,33 @@
...
+  'payment-interface'
+  'other'
+);
+
+ALTER TABLE "external-interface-description"
+-- Possible data content of the external interface
+  ADD COLUMN "data-content" interface_data_content[];
+
+
+-- Move data from external-interface-description/license_url to external-interface-description/external-interface/description
```
rckopteri: Really good description - iso peukku

database/src/main/resources/db/migration/V1_50__external_interfaces_description.sql
K1ll3rF0x: There is a lot of stuff to fix in addition to this, I was aware of this problem also. I'm in progress fixing the problems. If I could cancel the review request, that would be nice feature by GitHub :)

PR #285, opened 2018-02-01
15439774 Add search service list to admin panel

rckopteri:
# Added
 - New service search list for admin users
 - Let admin to open service but don't let them edit.

### DIFF
--- a/ote/src/cljs/ote/app/controller/transport_service.cljs
+++ b/ote/src/cljs/ote/app/controller/transport_service.cljs
@@ -402,3 +402,12 @@
               (dissoc s key)))
           service
           service-level-keys))
+
+(defn is-service-owner?
+  "Admin can see services that they don't own. So we need to know, if user is a service owner"
+  [app]
+  (let [service-operator-id (get-in app [:transport-service ::t-service/transport-operator-id])
+        is-owner? (some (fn [item]
+                          (if (= service-operator-id (get-in item [:transport-operator ::t-operator/id])) true false))
+                        (get app :transport-operators-with-services))]
+    (if is-owner? true false)))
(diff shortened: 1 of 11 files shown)


ote/src/cljs/ote/app/controller/transport_service.cljs
```diff
@@ -402,3 +402,12 @@
               (dissoc s key)))
           service
           service-level-keys))
+
+(defn is-service-owner?
+  "Admin can see services that they don't own. So we need to know, if user is a service owner"
+  [app]
+  (let [service-operator-id (get-in app [:transport-service ::t-service/transport-operator-id])
+        is-owner? (some (fn [item]
+                          (if (= service-operator-id (get-in item [:transport-operator ::t-operator/id])) true false))
```
K1ll3rF0x: Why if-clause? Equality-check would be enough.

ote/src/cljs/ote/app/controller/transport_service.cljs
```diff
@@ -402,3 +402,12 @@
...
+                        (get app :transport-operators-with-services))]
+    (if is-owner? true false)))
```
K1ll3rF0x: returning is-owner? would be enough.

rckopteri: some returns first instance that matches with the check.
So the is-owner? could be null. Which isn't nice.
I'll rename that variable.

K1ll3rF0x: Yes, that is true. is-owner? would be misleading because of the possible nil value. You could use a predicate any? or not-any? if you want to make sure that the return value is a boolean.

PR #286, opened 2018-02-01
154832365 Service search API

tatut:
# Added
- Support JSON output in service search backend calls

# Environment
- Add developer documentation

### DIFF
--- a/ote/src/clj/ote/services/service_search.clj
+++ b/ote/src/clj/ote/services/service_search.clj
@@ -118,7 +118,9 @@
      :results results
      :filter-service-count (count ids)}))
 
-(defn- parse-query-parameters [params]
+(defn- service-search-parameters
+  "Extract service search parameters from query parameters."
+  [params]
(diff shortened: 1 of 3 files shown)


tatut:
Contains JSON support for service search and some developer documentation.

ote/src/clj/ote/services/service_search.clj
```diff
@@ -118,30 +118,34 @@
      :results results
      :filter-service-count (count ids)}))
 
+(defn- parse-query-parameters [params]
```
rckopteri: Not a biggie, but might consider renaming this function.
It is kind of general now. It only parses parameters that are relevant to service search.
And in some occasions we have used to name functions by their return values.
So -> parsed-service-query-parameters. or something similar.

tatut: Absolutely right... and it has extra verbiage but is a pure function.

PR #287, opened 2018-02-02
External API url check

mmarkuss:
* Added
- External API address validity check


rckopteri:
Check conflicts

ote/src/cljs/ote/app/controller/transport_service.cljs
```diff
@@ -77,6 +77,10 @@
 (defrecord EnsureCsvFileResponse [response])
 (defrecord FailedCsvFileResponse [response])
 
+(defrecord EnsureUrl [url])
```
tatut: These are specific to external interface URLs... maybe they should be renamed to be something like EnsureExternalInterfaceUrl

PR #290, opened 2018-02-02
154832728 filter admin service list

rckopteri:
# Added
 - Published filter to admin service search


ote/src/clj/ote/services/admin.clj
```diff
@@ -42,23 +42,35 @@
...
+    (= :NO (:published-type query)) {::t-service/published? false}
+    :else nil))
 
 (defn- list-services
   "Returns list of transport-services. Query parameters aren't mandatory, but it can be used to filter results."
   [db user query]
-  (fetch db ::t-service/transport-service-search-result
+  (let [q (if (nil? (:query query))
+            nil
+            {::t-service/name (op/ilike (str "%" (:query query) "%"))})
```
K1ll3rF0x: Use (when (:query query) ...).

ote/src/clj/ote/services/admin.clj
```diff
@@ -42,23 +42,35 @@
...
-           {::t-service/name (op/ilike (str "%" query "%"))})
-         {:specql.core/order-by ::t-service/name}))
+         search-params
+         {:specql.core/order-by ::t-service/name})))
 
 (defn- list-services-by-operator [db user query]
-  (fetch db ::t-service/transport-service-search-result
+  (let [q (if (nil? (:query query))
+            nil
+            {::t-service/operator-name (op/ilike (str "%" (:query query) "%"))})
```
K1ll3rF0x: Use when.

ote/src/cljs/ote/views/admin.cljs
```diff
@@ -47,26 +50,39 @@
...
+      ]
 
+     [:div.row.col-md-2
+      [form-fields/field {:type :selection
+                          :label "Julkaistu?"
+                          :options published-types
+                          :show-option (tr-key [:admin-page :published-types])
+                          :update! #(e! (admin-controller/->UpdatePublishedFilter %))}
+       published-filter]
+      ]
```
K1ll3rF0x: Fix square bracket position.

ote/src/cljs/ote/views/admin.cljs
```diff
@@ -96,7 +112,7 @@
                [ui/table-row-column {:style {:width "20%"}} (tr [:enums :ote.db.transport-service/type (keyword type)])]
                [ui/table-row-column {:style {:width "20%"}} (tr [:enums :ote.db.transport-service/sub-type (keyword sub-type)])]
                [ui/table-row-column {:style {:width "5%"}} (if published? "Kyllä" "Ei") ]
-               [ui/table-row-column {:style {:width "15%"}}  (time/format-timestamp-for-ui created)]]))]]])])
+               [ui/table-row-column {:style {:width "15%"}}  (time/format-timestamp-for-ui created)]]))]]])]])
 
   )
```
K1ll3rF0x: Parenthesis position.

ote/src/clj/ote/services/admin.clj
```diff
@@ -42,23 +42,35 @@
 
 (defn- list-users [db user query]
   (nap-users/list-users db {:email (str "%" query "%")
-                            :name (str "%" query "%")}))
+                            :name  (str "%" query "%")}))
+(defn- published-search-param [query]
+  (cond
+    (or (nil? (:published-type query)) (= :ALL (:published-type query))) nil
```
tatut: wouldn't case be simpler?
```
(case (:published-type query)
  :YES {::t-service/published? true}
  :NO {::t-service/published? false}
  nil)
```

Even simpler if we just used `true`/`false` instead of `:YES`/`:NO` keywords

tatut: Even simpler: 
```
(when-let [pub (:published-type query)]
  {::t-service/published? 
   (case pub
    :YES true
    :NO false)})
```

rckopteri: :published-type can also be nil or :ALL. So there is 4 possibilities.

rckopteri: nil is possible, because material ui can't set default values for dropdownlist by default. It can be done by some sorcery, but I didn't want use anymore time to these admin view.

PR #295, opened 2018-02-05
154392247 translate geojson texts

mmarkuss:
* Fixed
- GeoJson view localization update for maximum parking time and availability time

### DIFF
--- a/ote/resources/public/language/en.edn
+++ b/ote/resources/public/language/en.edn
@@ -693,6 +693,14 @@
           "available-from" "Available from"
           "available-to" "Available to"
 
+          "maximum-stay" "Maximum parking time"
+          "seconds" "Seconds"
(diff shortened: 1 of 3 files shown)


ote/resources/public/language/en.edn
```diff
@@ -690,6 +690,17 @@
           "charging-points" "Kuvaus latauspisteistä"
           "parking-capacities" "Parkkipaikkojen määrä"
 
+          "available-from" "Available from"
+          "available-to" "Available to"
+
```
K1ll3rF0x: How are  :ote.db.transport-service/available-from-nil "Heti" :ote.db.transport-service/available-to-nil "Toistaiseksi" presented in the geojson?

mmarkuss: They are not presented at all.

PR #296, opened 2018-02-05
154440133 business-id report

rckopteri:
# Added
 - Business id report to admin panel


ote/src/clj/ote/services/admin.clj
```diff
@@ -73,6 +76,36 @@
          search-params
          {:specql.core/order-by ::t-service/operator-name})))
 
+(defn distinct-by [f coll]
+  (let [groups (group-by f coll)]
+    (map #(first (groups %)) (distinct (map f coll)))))
+
+(defn- business-id-report [db user query]
+  (let [services (if
```
tatut: Use `when` instead of `(if condition value nil)` pattern

ote/src/clj/ote/services/admin.clj
```diff
@@ -73,6 +76,36 @@
...
+                   nil)
+        operators (if
+                   (or
+                     (nil? (:business-id-filter query))
+                     (= :ALL (:business-id-filter query))
+                     (= :operators (:business-id-filter query)))
+                    (fetch-operator-business-ids db)
+                   nil)
+
+        report (cond
```
tatut: Why is this  `cond` structure necessary? just `(concat services operators)`

ote/src/cljs/ote/views/admin.cljs
```diff
@@ -49,6 +49,55 @@
...
+              [ui/table-header-column "Sähköposti"]]]
+          [ui/table-body {:display-row-checkbox false}
+           (doall
+             (for [{:keys [operator business-id phone email]} results]
+               [ui/table-row {:selectable false}
+                [ui/table-row-column operator]
+                [ui/table-row-column business-id]
+                [ui/table-row-column phone]
+                [ui/table-row-column email]
+                ]
```
tatut: Closing parens

ote/src/clj/ote/services/admin.clj
rckopteri: Changed

ote/src/clj/ote/services/admin.clj
rckopteri: Removed cond and replaced it with concat

ote/src/cljs/ote/views/admin.cljs
rckopteri: parens fixed

PR #304, opened 2018-02-08
154834176 fetch companies from url

rckopteri:
# Added
 - Save external companies to service-company table
 - Enable possibility to tell users how many companies can't be saved
 - Add business-id validator

# Environment
 - Added service-company table


ote/src/clj/ote/services/external.clj
```diff
@@ -2,13 +2,26 @@
...
+            [specql.core :as specql]))
+
+(defn valid-csv-header?
+  "Ensure that there is at least 2 elements in header"
+  [header]
+  (>= (count header) 2))
+
+(defn valid-business-id? [value]
+  (let [pattern  #"\d{7}-\d"]
+  (if  (re-matches pattern value)
```
tatut: Avoid `(if condition true false)` pattern.

Often you can just use `condition` by itself or wrap it with `boolean` to force a `true`/`false` value.

```
(defn valid-business-id? [business-id]
  (boolean (re-matches #"\d{7}-\d" business-id)))
```

ote/src/clj/ote/services/external.clj
```diff
@@ -20,25 +33,46 @@
...
-       (rest csv-data)))
+  (let [headers (first csv-data)
+        valid-header? (valid-csv-header? headers)
+        parsed-data (when valid-header?
+                      (map (fn [[business-id name]]
+                             {::t-service/business-id business-id
+                              ::t-service/name name})
+                            (rest csv-data)))
+        validated-data (filter #(valid-business-id? (::t-service/business-id %)) parsed-data)
+        filtered-data  (map #(dissoc % :remove) validated-data )]
```
tatut: Why the `#(dissoc % :remove)` I don't see a key named `:remove` in the parsed data, perhaps it is stale

ote/src/clj/ote/services/external.clj
```diff
@@ -20,25 +33,46 @@
...
+    {:result filtered-data
+     :failed-count (- (count parsed-data) (count filtered-data))}))
 
-(defn- check-csv
+(defn save-companies
+  "Save business-ids, company names to db"
+  [db data]
+   (let [data (if (nil? (::t-service/id data))
```
tatut: We have `ote.db.modification` ns and the `created` and `modified` columns should be mapped to keywords in that namespace as is done with other tables.

Also use `with-modification-fields` helper that conditionally adds creation/modification.

ote/src/clj/ote/services/external.clj
```diff
@@ -20,25 +33,46 @@
...
+               (assoc data ::t-service/created (java.sql.Timestamp. (System/currentTimeMillis)))
+               ;; And modified for the rest
+               (assoc data ::t-service/modified (java.sql.Timestamp. (System/currentTimeMillis))))]
+    (specql/upsert! db ::t-service/service-company data)))
+
+(defn check-csv
   "Fetch csv file from given url, parse it and return status map that contains information about the count of companies
   in the file."
-  [url-data]
+  [db url-data]
```
tatut: `db` parameter is not used, remove it

ote/src/clj/ote/services/transport.clj
```diff
@@ -247,17 +248,41 @@
...
+  [db transport-service]
+    (specql/delete! db ::t-service/service-company {::t-service/transport-service-id (::t-service/id transport-service)}))
+
+(defn- save-external-companies
+  "Service can contain an url that contains company nmes and business-id. Sevice can also contain an imported csv file
+  with company names and business-ids."
+  [db transport-service]
+  (let [current-data (first (fetch db ::t-service/service-company (specql/columns ::t-service/service-company)
+                            {::t-service/transport-service-id (::t-service/id transport-service)}))
+        companies (into [] (:companies (external/check-csv db {:url (get transport-service ::t-service/companies-csv-url)
```
tatut: Destructure `::t-service/companies-csv-url` and `::t-service/id` to make this clearer.

ote/src/clj/ote/services/transport.clj
```diff
@@ -247,17 +248,41 @@
...
+                                                                            :ts-id (get transport-service ::t-service/id)})))
+        new-data (if (empty? current-data)
+                      {::t-service/companies            companies
+                       ::t-service/transport-service-id (get transport-service ::t-service/id)
```
tatut: Use destructured value (see above comment)

ote/src/clj/ote/services/transport.clj
```diff
@@ -247,17 +248,41 @@
...
+                       ::t-service/source               "URL"}
+                      (assoc current-data ::t-service/companies companies))]
+
+    (external/save-companies db new-data)
+  ))
```
tatut: closing parens

ote/src/clj/ote/services/transport.clj
```diff
@@ -247,17 +248,41 @@
...
                        external-interfaces ::t-service/external-interfaces
+                       service-company ::t-service/service-company
                        :as data}]
   ;(println "DATA: " (pr-str data))
   (let [service-info (-> data
                          (modification/with-modification-fields ::t-service/id user)
                          (dissoc ::t-service/operation-area)
                          floats-to-bigdec
-                         (dissoc ::t-service/external-interfaces))
+                         (dissoc ::t-service/external-interfaces)
```
tatut: Combine these two dissocs with the above, 1 `dissoc` call can take multiple keys.

ote/src/cljc/ote/db/transport_service.cljc
```diff
@@ -141,3 +142,6 @@
 
 (def rental-transportable-aid
   [:wheelchair])
+
+(def service-company-soure
+  [:FILE :CSV :URL])
```
tatut: ALL-CAPS is unusual style for keywords, lowercase

ote/src/cljc/ote/db/transport_service.cljc
```diff
@@ -67,7 +67,8 @@
 
   ["external_interface_search_result" ::external-interface-search-result]
   ["transport_service_search_result" ::transport-service-search-result
-   ote.db.modification/modification-fields])
+   ote.db.modification/modification-fields]
+  ["service_company" ::service-company])
```
tatut: Map modification columns to `ote.db.modification` ns

ote/src/clj/ote/services/external.clj
rckopteri: I thought of using with-modification-fields function. But in it's current form it also requires to table to have a user. service-company table is mainly used with our scheduler so I'm not adding user-id to it.
So I thought that there is no reason to map those fields any differently as any other field in that table.

ote/src/clj/ote/services/external.clj
rckopteri: yep. good catch, your mapping magic removed the need for :remove key.

ote/src/clj/ote/services/transport.clj
rckopteri: changed

ote/src/clj/ote/services/transport.clj
rckopteri: changed

ote/src/clj/ote/services/transport.clj
rckopteri: fixed

ote/src/cljc/ote/db/transport_service.cljc
rckopteri: Do I have to?

tatut: Somebody has to do it. We don't want multiple definitions for the same columns, it is confusing for future maintenance when sometimes the same column is in one namespace and other times in another.

rckopteri: done

PR #305, opened 2018-02-08
Add naughty generator for cypress, add naughty form fill spec.

K1ll3rF0x:
# Environment
* Add naughty string generator utility for cypress.
* Create custom .typeRaw(val) command for cypress (enables injecting any string data into input fields).
* Add naughty form fill test that tries to crash our taxi form by filling it with evilness.


cypress/support/generators/index.js
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        return generate(len, `${acc}${sampleArr(srcArr)}`);
+    };
+
+    return generate(len, sampleArr(srcArr));
+};
+
+export const genNaughtyString = (len) => {
+    const generate = (len, acc) => {
```
tatut: arent these `const generate` utils nearly the same in `genString` and `genNaughtyString`?  Can we make one utility to use in both

K1ll3rF0x:
Yes, good catch. A accidentally copy pasted my work from repl into a wrong function.
________________________________________

PR #310, opened 2018-02-12
154872430 Migration to alter collation of transport operator/service name

tatut:
# Environment
- Change collation of operator/service name (for better result ordering)


rckopteri:
I'm afraid that we don't remember to add collate_FI parameter to all text columns that we will add in the future. So it would be better to change the database collation.

tatut:
Not possible (at least I don't know how to do it).

The only way AFAICT is to fully recreate the database, which I don't want to do and would require downtime.

PR #332, opened 2018-02-23
154918168 ulkoisten rajapintojen formaatti  ja lisenssi kenttien uudistaminen

K1ll3rF0x:
# Added
* Added autocomplete and chip-input form fields.
* External interface form group changes:
  * Using autocomplete field type for license field. Added some license suggestions.
  * Using chip-input field type for format field. Added some response file format suggestions.

# Datamodel
* Format attribute/column is now type of TEXT[].

### DIFF
--- a/ote/src/cljs/ote/views/transport_service_common.cljs
+++ b/ote/src/cljs/ote/views/transport_service_common.cljs
@@ -139,6 +139,7 @@
                        :tooltip (tr [:form-help :external-interfaces-tooltips :format])
                        :width "15%"
                        :full-width? true
+                       :full-width-input? true
                        :required? true}
                       {:name ::t-service/license
                        :type :autocomplete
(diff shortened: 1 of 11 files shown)


ote/src/cljs/ote/views/transport_service_common.cljs
```diff
@@ -151,16 +151,23 @@
                        :width "5%"
                        }
                       {:name ::t-service/format
-                       :type :string
+                       :type :chip-input
+                       :open-on-focus? true
+                       :suggestions ["GTFS" "Kalkati.net" "SIRI" "NeTEx" "GeoJSON" "JSON" "CSV"]
```
rckopteri: Could be in separate array and not hard coded here. Would help if these are moved to database and administrated by admin in some day.

K1ll3rF0x: Yes, currently both licenses and format suggestions are hardcoded. They can be added into database if we decide they should be admistrated.

PR #352, opened 2018-03-08
155062911 Fix paragraphs in external interface modal

rckopteri:
# Fixed
 - Fix paragraphs in external interface modal
 by removing margins only from the first p element.

Toni, did I do well?

### DIFF
--- a/ote/resources/public/css/styles.css
+++ b/ote/resources/public/css/styles.css
@@ -35,7 +35,7 @@ h2 {
 
 /* Rendered translations are wrapped with <p> by Marked JS library.
    Remove the margins of the first P-element */
-.translation-markdown > p {
+.translation-markdown > :first p {
     margin-top: 0;
     margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


ote/resources/public/css/styles.css
```diff
@@ -35,7 +35,7 @@ h2 {
 
 /* Rendered translations are wrapped with <p> by Marked JS library.
    Remove the margins of the first P-element */
-.translation-markdown > p {
+.translation-markdown > :first p {
```
K1ll3rF0x: This should not even work. :first is used only with @page rule and it does not do anything here.
The correct solution would be:
```css
.translation-markdow p:only-child {
     margin-top: 0;
     margin-bottom: 0;
}
```
This removes margins from `<p>`, only if `<p>` is the only child of .translation-markdown element. Otherwise, if `<p>` has sibling elements, it has normal margins.

PR #356, opened 2018-03-12
155891093 Save sea route to database

tatut:
# Added
- Save sea route to database


ote/src/clj/ote/services/routes.clj
```diff
@@ -54,13 +57,23 @@
                                       routes)})
          operators)))
 
+(defn save-route [nap-config db user route]
+  (println "************ route " (pr-str route))
```
mmarkuss: remove print

PR #358, opened 2018-03-13
155933397 add cancel and links to route view

rckopteri:
# Added
 - Cancel button to sea route wizard
 - Links to wizard steps to make it more usable when editing route

### DIFF
--- a/ote/src/cljs/ote/app/controller/route.cljs
+++ b/ote/src/cljs/ote/app/controller/route.cljs
@@ -42,6 +42,7 @@
 
 ;; Save route to database
 (defrecord SaveToDb [])
+(defrecord CancelRoute [])
 (defrecord SaveRouteResponse [response])
 (defrecord SaveRouteFailure [response])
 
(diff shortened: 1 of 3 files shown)


ote/src/cljs/ote/views/route.cljs
```diff
@@ -64,12 +67,16 @@
...
+     [ui/stepper {:active-step i
+                  :connector (r/as-element [ic/navigation-arrow-forward])}
       (doall
-       (for [{label :label} wizard-steps]
-         ^{:key label}
-         [ui/step
-          [ui/step-label label]]))]
+        (for [{label :label current-step :name} wizard-steps]
+          ^{:key label}
+          [ui/step (when (rc/validate-all-steps route current-step wizard-steps)
```
tatut: Calculate validity once and put it in the `for` :let, not it is repeated twice: once for style and once for click handler

ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -260,3 +266,13 @@
                    (every? #(and (time/valid-time? (::transit/departure-time %))
                                  (time/valid-time? (::transit/arrival-time %))) other-stops))))
           trips))
+
+(defn validate-all-steps
```
tatut: I think a term like "previous" would be better as it does not validate "all" steps.

ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -229,7 +230,12 @@
   (process-event [{response :response} app]
     (.error js/console "Save route failed:" (pr-str response))
     (assoc app
-           :flash-message-error "Reitin tallennus epäonnistui")))
+      :flash-message-error "Reitin tallennus epäonnistui"))
```
mmarkuss: Add text to resource file.

rckopteri: Yes, later.
I think that we can use in this poc only finnish and translate all text at once.

PR #359, opened 2018-03-13
155847048 operation area drawing

mmarkuss:
# Fixed
- Hide operation drawing tools by default.
- Let user select drawing tools for primary and secondary operation areas.

### DIFF
--- a/ote/src/cljs/ote/app/controller/place_search.cljs
+++ b/ote/src/cljs/ote/app/controller/place_search.cljs
@@ -10,6 +10,7 @@
 
 (defrecord SetPrimaryPlaceName [name])
 (defrecord SetSecondaryPlaceName [name])
+(defrecord SetDrawControl [show? primary?])
 (defrecord AddPlace [id primary?])
 (defrecord FetchPlaceResponse [response place])
 (defrecord RemovePlaceById [id])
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/views/place_search.cljs
```diff
@@ -186,8 +194,30 @@
                          :on-update-input #(e! (ps/->SetSecondaryPlaceName %))
                          :search-text (or (:secondary-name place-search) "")
                          :on-new-request #(e! (ps/->AddPlace (aget % "id") false))}]]
+
      [:div {:style {:width "70%"}}
-       [places-map e! (:results place-search)]]]))
+       [places-map e! (:results place-search) (:show? place-search)]
+
+      [:span
```
rckopteri: I prefer divs

ote/src/cljs/ote/app/controller/place_search.cljs
```diff
@@ -66,6 +67,13 @@
   (process-event [{name :name} app]
     (search (assoc-in app [:place-search :secondary-name] name) false))
 
+  SetDrawControl
+  (process-event [{show? :show? primary? :primary?} app]
+    (if (nil? primary?)
+      (assoc-in (update-in app [:place-search] dissoc :primary?) [:place-search :show?] show?)
```
rckopteri: I don't know if this is ok, but I would suggest to change this like this:
(-> app
 ( update-in ...)
(assoc-in ....)
(assoc-in ...)
(assoc-in ...)

ote/src/cljs/ote/views/place_search.cljs
mmarkuss: I'll leave this span which Toni recommended using here.

ote/src/cljs/ote/app/controller/place_search.cljs
mmarkuss: Fixed.

PR #363, opened 2018-03-15
155616209 stop exceptions

rckopteri:
# Added
 - Pick up and drop off exceptions


ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -46,6 +47,19 @@
 (defrecord SaveRouteResponse [response])
 (defrecord SaveRouteFailure [response])
 
+(defn- update-stop-by-idx [route stop-idx update-fn & args]
+  (let [data (mapv
```
mmarkuss: Small thing: is update-fn really needed?

PR #379, opened 2018-03-19
155971065 sea traffic wizard state fixes

K1ll3rF0x:
# Fixes
* Fixed problems with related to sea traffic wizard state initialization
  * Wizard state does not reset when figwheel updates the code.
  * Wizard state does not reset when jumping between steps.
  * Reset "Vuorot"-step state if editing routes.
* Fix a problem that caused wizard save to fail while wizard calendar editor is open.

# Changes
* Combine LoadStops and InitRoute events.

(#155971065, #156015288)


ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -58,25 +58,26 @@
                   stops))))
 
 (extend-protocol tuck/Event
-  LoadStops
+  InitRoute
```
rckopteri: I know that this seems to be the way to go right now, but it is actually good thing that these functions are separated.
When saved route is loaded, those route points needs to be loaded without reseting loaded data (from db).

tatut: The determination of what needs to done differently in editing vs new scenarios should be done in the controller. So the InitRoute is a good name for that, but could be split by making some functions it calls.

PR #380, opened 2018-03-20
155972205 Add custom stops to route

tatut:
# Added
- Add stops (finnish ports) to database
- Support adding custom stops to route


database/src/main/resources/db/migration/V1_69__finnish_ports.sql
```diff
@@ -0,0 +1,97 @@
+CREATE TABLE finnish_ports (
+ id SERIAL PRIMARY KEY,
+ code TEXT,
+ name TEXT,
+ location GEOMETRY,
+ created timestamp with time zone DEFAULT NOW(),
+ "created-by" TEXT REFERENCES "user" (id)
```
rckopteri: Add one column to rule them all.
We need to know if column is added by user and approved by admin.
Not sure when those admins will approve anything, but it won't hurt if we have database support for it already.

I would suggest that all those ports that are from valid source are indicated something like
official_finnish_port = true
and all unvalidated custom ports are indicated something like
validated = false.
So the table could be like
id SERIAL PRIMARY KEY,
code TEXT,
name TEXT,
location GEOMETRY,
created timestamp with time zone DEFAULT NOW(),
"created-by" TEXT REFERENCES "user" (id)
official_finnish_port BOOL,
validated BOOL

tms.

ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -97,6 +108,81 @@
         (assoc-in [:route ::transit/trips] [])
         (assoc-in [:route ::transit/service-calendars] [])))
 
+  AddCustomStop
+  (process-event [{id :id} {route :route :as app}]
+    ;; Add stop to current stop sequence
+    (let [{feature :geojson :as custom-stop}
+          (first (keep #(when (= (:id %) id) %) (:custom-stops route)))]
+      (.log js/console "custom-stop:" (pr-str custom-stop))
```
rckopteri: No need for this

ote/src/cljs/ote/app/controller/route.cljs
```diff
@@ -97,6 +108,81 @@
...
+                     (if (= id stop-id)
+                       (update stop :geojson
+                               #(-> %
+                                    js->clj
+                                    (assoc :geometry (get (js->clj geojson) "geometry"))
+                                    clj->js))
+                       stop)))
+        (update-in [:route ::transit/stops] (flip mapv)
+                   (fn [{::transit/keys [custom code] :as stop}]
+                     (.log js/console id " == " code "?" (= code id))
```
rckopteri: Can be removed

ote/src/cljs/ote/views/route/stop_sequence.cljs
```diff
@@ -17,29 +20,93 @@
...
+       [leaflet/Map {:ref "stops-map"
+                     :center #js [65 25]
+                     :zoomControl true
+                     :zoom 5}
+        (leaflet/background-tile-map)
+        (when-let [stops (:stops route)]
+          [leaflet/GeoJSON {:data stops
+                            :style {:color "green"}
+                            :pointToLayer (partial stop-marker e!)}])
+        #_(for [s (:custom-stops route)]
```
rckopteri: commented code can be removed

database/src/main/resources/db/migration/V1_69__finnish_ports.sql
tatut: I don't want to add columns prematurely, we can already determine initial data because it has no creator.

Will ALTER TABLE when we get new requirements.

PR #381, opened 2018-03-20
155971785 copy calendar

mmarkuss:
# Added
- Adding a new trip is now allowed only after an user has defined a calendar schedule for the first trip.

### DIFF
--- a/ote/src/cljs/ote/views/route/trips.cljs
+++ b/ote/src/cljs/ote/views/route/trips.cljs
@@ -137,7 +137,6 @@
   (let [stop-sequence (::transit/stops route)
         stop-count (count stop-sequence)
         trips (::transit/trips route)]
-
     [:div.route-times
      [:table {:style {:text-align "center"}}
       [route-times-header stop-sequence]
@@ -149,7 +148,8 @@
                           :update! #(e! (rc/->NewStartTime %))} (:new-start-time route)]
       [ui/raised-button {:style {:margin-left "5px"}
                          :primary true
-                         :disabled (time/empty-time? (:new-start-time route))
+                         :disabled (or (time/empty-time? (:new-start-time route))
+                                        (empty? (first (:ote.db.transit/service-calendars route))))
                          :on-click #(e! (rc/->AddTrip))
                          :label "Lisää vuoro"}]]]))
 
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/views/route/trips.cljs
```diff
@@ -136,20 +138,25 @@
 (defn trips-list [e! route]
   (let [stop-sequence (::transit/stops route)
         stop-count (count stop-sequence)
-        trips (::transit/trips route)]
-
+        trips (::transit/trips route)
+        empty-calendar? (empty? (first (:ote.db.transit/service-calendars route)))]
```
K1ll3rF0x: Use ::transit/service-calendars

ote/src/cljs/ote/views/route/trips.cljs
```diff
@@ -136,20 +138,25 @@
...
     [:div.route-times
      [:table {:style {:text-align "center"}}
       [route-times-header stop-sequence]
       [:tbody
        (doall (map-indexed (partial trip-row e! stop-count) trips))]]
+
+     (when empty-calendar?
+       [:div {:style {:margin-left "25px" :margin-right "25px" :margin-top "10px"}}
```
K1ll3rF0x: Another opinion; the help box looks better if you remove margin-left and margin-right.

ote/src/cljs/ote/views/route/trips.cljs
```diff
@@ -136,20 +138,25 @@
...
-
+        trips (::transit/trips route)
+        empty-calendar? (empty? (first (:ote.db.transit/service-calendars route)))]
     [:div.route-times
      [:table {:style {:text-align "center"}}
       [route-times-header stop-sequence]
       [:tbody
        (doall (map-indexed (partial trip-row e! stop-count) trips))]]
+
+     (when empty-calendar?
```
K1ll3rF0x: This is an opinion, but wouldn't it be better to keep the infobox always visible? For example, I forgot the instructions after I opened the calendar open and went back to trips editor.

ote/src/cljs/ote/views/route/trips.cljs
mmarkuss: Fixed.

ote/src/cljs/ote/views/route/trips.cljs
mmarkuss: Fixed.

ote/src/cljs/ote/views/route/trips.cljs
mmarkuss: I'll leave this as it is now based on UI specs but we can remove it later if needed.

PR #382, opened 2018-03-20
155935823 Allow round trips.

mmarkuss:
# Fixed
- Round trips are allowed routes.


tatut:
Also check GTFS export so that it still works

PR #386, opened 2018-03-21
156003906 add translation tags

rckopteri:
# Added
 * Add translation tags
 * Take english translations into use


ote/src/cljs/ote/ui/service_calendar.cljs
```diff
@@ -85,18 +85,18 @@
...
-    4 "Huhti"
-    5 "Touko"
-    6 "Kesä"
-    7 "Heinä"
-    8 "Elo"
-    9 "Syys"
-    10 "Loka"
-    11 "Marras"
-    12 "Joulu"))
+    1 (tr [:route-wizard-page :short-month-names :JAN])
```
tatut: This is a common UI widget, I think we should use a more "common" place for the translations, not under `:route-wizard-page`.

Better yet, I think we can use JS Dates formatting:
`(.toLocaleString (doto (js/Date.) (.setMonth 0)) "fi" #js {:month "short"})` (returns "tammi")

ote/src/cljs/ote/ui/service_calendar.cljs
```diff
@@ -85,18 +85,18 @@
 (defn- month-name [month]
   ;; FIXME: use translation
```
tatut: Remove FIXME

PR #394, opened 2018-03-23
156242543 Kalkati2gtfs

K1ll3rF0x:
# Added
* AWS Lambda (test) API for converting kalkati zip files to gtfs zips.
Use it like:
```bash
$ wget --post-data '"kalkati-api-url/all.zip"' https://<censored>.execute-api.eu-central-1.amazonaws.com/deployed/kalkati_to_gtfs -O gtfs.zip
```


tatut:
Nice! 

I think we need to add security group configuration so that only our ec2 instances can call this lambda (so as to not leave an open proxy).

K1ll3rF0x:
Yes, absolutely. I left the proxy open for easy testing. Also, the lambda is running currently on our test aws environment (to prevent messing up anything while developing). We'll have to move it into production after testing.

PR #399, opened 2018-03-27
156151334 own ports not shown in map

rckopteri:
# Fixed
 * Own port map marker was missing if route wizard step was changed

# Added
 * Save and exit button to route wizard


tatut:
I'll investigate the Leaflet.Draw integration, let's not merge this (yet).

PR #400, opened 2018-03-27
156151334 Fix disappearing custom stops bug

tatut:
# Fixed
- Custom stops no longer disappear when going to another wizard step and coming back to the map


tatut:
Please test in sandbox environment.

rckopteri:
Its now running in: http://ec2-52-59-215-92.eu-central-1.compute.amazonaws.com/
Custom stops don't disappear from map when changing route wizard step but when I go to Trip step and come back new custom stops don't appear to the map anymore.

PR #406, opened 2018-04-03
156262985 Sea route UX improvements

tatut:
# Added
- Tooltips for sea route icon buttons

# Changed
- Styling improvements to sea route form


ote/src/cljs/ote/views/route/service_calendar.cljs
```diff
@@ -65,7 +65,9 @@
        "Ajopäiväkalenteri"]
```
rckopteri: Do we still have some untranslated texts?

tatut: I'll add it to translations

PR #413, opened 2018-04-05
156461755 Service calendar dialog

tatut:
# Changed
- Trip service calendar is now in a dialog

### DIFF
--- a/ote/resources/public/language/en.edn
+++ b/ote/resources/public/language/en.edn
@@ -664,7 +664,7 @@
   :basic-info-available-to "Valid to"
 
   :route-calendar-name "Reittikalenteri"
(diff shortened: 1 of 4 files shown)


ote/resources/public/language/en.edn
```diff
@@ -664,7 +664,7 @@
   :basic-info-available-to "Valid to"
 
   :route-calendar-name "Reittikalenteri"
-  :route-calendar-group-name "Driving days calendar"
+  :route-calendar-group-name ["Service calendar (Departs from " :departure-stop " at " :departure-time ")"]
```
rckopteri: In Finnis we have "Liikennöinti", in Swedish "trafiken" but in english "service". Should it be "Traffic" of "route"?

PR #420, opened 2018-04-09
156493140 calendar validation

rckopteri:
# Added
 * Added validation to calendar rules. Now user cannot select which week days route is transiting if from/to dates are not set.
 * Added  validation to calendar notification so that missing rule dates are not indicated if they are not missing. Previously user could add few transit days by hand and validator didn't notice that.

### DIFF
--- a/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
+++ b/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
@@ -615,3 +615,20 @@
 
 (defn valid-name [route]
   (if (empty? (get route ::transit/name)) false true))
+
+(defn valid-calendar-rule-dates? [data]
+  (every?
+    (fn [rule]
+      (if (and
+            (not (str/blank? (get rule ::transit/from-date)))
+            (not (str/blank? (get rule ::transit/to-date)))
+            )
+        true false))
+    (::transit/service-rules data)))
+
+(defn valid-calendar-from-tro-dates? [data]
+  (if (and
+        (not (str/blank? (get data ::transit/from-date)))
+        (not (str/blank? (get data ::transit/to-date)))
+        )
+    true false))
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -615,3 +616,20 @@
 
 (defn valid-name [route]
   (if (empty? (get route ::transit/name)) false true))
+
+(defn valid-calendar-rule-dates? [data]
+  (every?
+    (fn [rule]
+      (if (and
```
tatut: remove `if`,  `and` will return true/false

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -615,3 +616,20 @@
...
+            (not (str/blank? (get rule ::transit/from-date)))
+            (not (str/blank? (get rule ::transit/to-date)))
+            )
+        true false))
+    (::transit/service-rules data)))
+
+(defn valid-calendar-from-tro-dates? [data]
+  (if (and
```
tatut: remove `if`, formatting

ote/src/cljs/ote/ui/form_fields.cljs
```diff
@@ -733,26 +733,29 @@
         [buttons/save {:on-click #(update! (conj (or data []) {}))
                        :label add-label
                        :label-style style-base/button-label-style
-                       :disabled (values/effectively-empty? (last data))}]])]))
+                       :disabled (if add-label-validate
+                                   (not (add-label-validate (last data)))
```
tatut: I don't understand the name  `add-label-validate`, it doesn't add a label but determines disabled state

rckopteri: add-label should be "add-new-row-button" or something.
Then add-label-validate would be "add-new-row-button-validate". But you wouldn't like that?
I will change it to :add-label-disabled?

PR #427, opened 2018-04-10
156495547 Fix route trips help box visibility issue

K1ll3rF0x:
# Fixed
* Hide route trips help box only when the first calendar is valid.


ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -586,11 +586,11 @@
        transport-operator-id))
 
 (defn valid-calendar? [trip-calendar]
-  (if (or
+  (not (or
```
rckopteri: nice!

PR #429, opened 2018-04-11
Pre notice

rckopteri:
# Added
* flag for pre-notices (won't show up in production)
* list page for pre notices (not ready)
* form for pre notices (not ready)


ote/src/cljs/ote/views/pre_notices/pre_notice.cljs
```diff
@@ -0,0 +1,213 @@
...
+    [form-fields/field
+     {:label       (tr [:field-labels :select-transport-operator])
+      :name        :select-transport-operator
+      :type        :selection
+      :show-option ::t-operator/name
+      :update!     #(e! (pre-notice/->SelectOperatorForNotice %))
+      :options     operators
+      :auto-width? true}
+     operator]]
+   [:div.row.col-xs-12.col-sm-12.col-md-8
```
tatut: why not use `ote.ui.common/table2` to show info, do we need form fields for read-only data... we could just show the information without having editable components that are disabled

rckopteri: Jussi oli piirtänyt nuo disabloidut kentät tuonne. Minusta tuon ois voinut tehdä nätimminkin.

PR #436, opened 2018-04-12
156533232 pre notice listing for transit authority

tatut:
# Added
- Pre noticy listing for transit authority users

### DIFF
--- a/ote/resources/public/language/en.edn
+++ b/ote/resources/public/language/en.edn
@@ -690,7 +690,9 @@
 
  :pre-notice-list-page
  {:header-pre-notice-list "Omat 60pv ilmoitukset"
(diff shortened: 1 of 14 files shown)


ote/resources/public/language/en.edn
```diff
@@ -690,7 +690,9 @@
 
  :pre-notice-list-page
  {:header-pre-notice-list "Omat 60pv ilmoitukset"
+  :header-authority-pre-notice-list "Säännöllisen henkilöliikenteen muutosilmoitukset"
```
rckopteri: I prefer own section for every page.
And I know that many of these texts are the same, but still. It might be easier to handle when texts are separated to different sections by page.
-> :pre-notice-ely-list-page {}

PR #438, opened 2018-04-13
156724067 new header

rckopteri:
# Changed
* Update 60 days pre notice header texts
* Create header component and take it in use in list pages
* Change pre notice list pages background to gray


ote/src/cljs/ote/ui/list_header.cljs
```diff
@@ -0,0 +1,10 @@
+(ns ote.ui.list_header
```
mmarkuss: Smallie: change underscore to hyphen.

PR #439, opened 2018-04-13
156687806 60 day route information

mmarkuss:
# Added
* Pre-notice route change information


ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -39,4 +42,10 @@
    (POST "/pre-notice" {form-data :body
                         user      :user}
          (http/transit-response
-          (save-pre-notice db user (http/transit-request form-data))))))
+          (save-pre-notice db user (http/transit-request form-data))))
+   (GET "/pre-notices/regions" {}
+        (http/transit-response (fetch-regions db)))
```
rckopteri: I suggest that you wrap this db query to function, because we might need to do something about those returned values.
E.g. translate them or something

tatut: I don't understand. If we need to do something to the result, we can change the code then.

PR #440, opened 2018-04-13
156752682 happy test case 60days form

rckopteri:
# Added
* HappyCase unit test for 60days form


ote/test/clj/ote/db/service_generators.clj
```diff
@@ -6,12 +6,27 @@
...
             [ote.db.transport-service :as t-service]
             [ote.db.common :as common]
             [ote.time :as time]
             [clojure.string :as str]
             [ote.db.generators :as generators]))
 
+(def gen-effective-dates
+  (gen/hash-map
+    ::transit/effective-date generators/gen-effective-date
+    ::transit/effective-date-description generators/gen-text))
```
K1ll3rF0x: Why not use the naughty string generator instead of creating a new text generator?

rckopteri: Yep. good idea

PR #445, opened 2018-04-17
156838441 validate 60days form

rckopteri:
# Added
* Possibility to add attachments to 60days form
   
# Fixed
* Refresh operator problem in sea routes form


ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -16,14 +16,31 @@
 
 (defqueries "ote/services/pre_notices/regions.sql")
 
+(defn- parse-attachments
+  [attachments db user]
+  (let [saved-attachments (mapv
+                            #(first (specql/fetch db ::transit/pre-notice-attachment
+                                           (specql/columns ::transit/pre-notice-attachment)
+                                                  (op/and
+                                                  {::transit/attachment-file-name (::transit/attachment-file-name %)
```
tatut: Get attachment by `id`, not name... you can reasonably expect there to be files with the same name. Also id, being the primary key, will always be indexed.

ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -16,14 +16,31 @@
...
+                            attachments)]
+    saved-attachments))
+
 (defn get-operator-pre-notice [db user id]
   "Get singular operator pre-notice by id"
   (http/no-cache-transit-response
-    (first (specql/fetch
+    (let [pre-notice (first (specql/fetch
              db ::transit/pre-notice
              (specql/columns ::transit/pre-notice)
```
tatut: Could you just join the attachments and let specql handle it? There are no other collections being fetched.

ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -36,10 +53,19 @@
...
+               n (-> notice
+                     (dissoc :attachments)
+                     (modification/with-modification-fields ::transit/id user))
+               saved-notice (specql/upsert! db ::transit/pre-notice n)
+               notice-id (::transit/id saved-notice)
+               attachments (parse-attachments attachment-names db user)
+               ments (mapv #(assoc % ::transit/pre-notice-id notice-id) attachments)]
+           (log/debug "Save notice: " saved-notice)
+           (log/debug "Attachments: " attachments)
+           (doall
```
tatut: I presume this is done for the side-effect, not the return value.
It is clearer to use `doseq` instead of `doall` and `mapv`.

Instead of upserting, it's better to `update!` (we know that the attachment exists, we are just changing one field).

tatut: what are `ments`?

ote/src/cljs/ote/app/controller/pre_notices.cljs
```diff
@@ -11,7 +11,8 @@
 
 
 (defn valid-notice? [notice]
-  true)
+  (and (not (form/errors? notice))
```
tatut: this is the same as `form/valid?`

ote/src/cljs/ote/app/controller/pre_notices.cljs
```diff
@@ -212,3 +213,12 @@
...
                :on-failure (tuck/send-async! ->ServerError)})
   (update app :pre-notice-dialog dissoc :new-comment))
+
+(define-event UploadResponse [response]
+  {:path [:pre-notice :attachments]}
+  (conj (or (vec (butlast app)) []) response))
+
+(define-event UploadAttachment [input]
+  {}
+  (comm/upload! "pre-notice/upload" input {:on-success (tuck/send-async! ->UploadResponse)})
```
tatut: Handle `:on-failure` case with `->ServerError`

PR #447, opened 2018-04-18
156794815 help link button

rckopteri:
# Added
* Link to sea route editor user guide

### DIFF
--- a/ote/src/cljs/ote/ui/buttons.cljs
+++ b/ote/src/cljs/ote/ui/buttons.cljs
@@ -1,8 +1,10 @@
 (ns ote.ui.buttons
   (:require [cljs-react-material-ui.reagent :as ui]
+            [ote.localization :refer [tr tr-key]]
             [cljs-react-material-ui.core :refer [color]]
             [stylefy.core :as stylefy]
-            [ote.style.base :as style-base]))
+            [ote.style.base :as style-base]
+            [ote.app.controller.front-page :as fp-controller]))
 
 
 (defn- button-container [button]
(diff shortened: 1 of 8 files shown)


ote/src/cljs/ote/ui/buttons.cljs
```diff
@@ -27,3 +29,11 @@
                     :font-size "12px"
                     :font-weight "bold"}} opts)
     label]])
+
+(defn open-link [e! url]
```
tatut: could we use the common `linkify` utility instead of this new button?

K1ll3rF0x: Yes, please.

rckopteri: We could, but ui spec says that it must be a flat-button

K1ll3rF0x: Well, you could do it like this:
```cljs
[ui/flat-button
  {:label (r/as-element 
               (common/linkify "http://nonsense" "Go!" {:target "_blank"}))
  :primary true}]
```

rckopteri: That won't work. You can press the button and nothing happens if mouse pointer is not on top of label.

K1ll3rF0x: It turns out, you can actually do it like this:
```cljs
 (common/linkify "http://nonsense" 
  [ui/flat-button {:label "Go! :primary true}]
   {:target "_blank"})
```
The a-element will properly wrap the "button" element and the whole button is then interpreted as a link.

rckopteri: This is why I love these reviews. Thanx! That linkify works marvellously.
Changes coming up.

PR #449, opened 2018-04-20
156836192 Add missing fields to authority pre notice view

tatut:
# Added
- Operator, regions and attachments are now visible in authority pre notice view


ote/src/cljs/ote/ui/common.cljs
```diff
@@ -29,9 +29,18 @@
      (if-not url
        [:span]
        ;; Lazy check if url has a protocol, or if it is a relative path
-       (let [url (if (re-matches #"^(\w+:|.?.?/).*" url)
+       (let [url (cond
```
rckopteri: Wouldn't hurt if we had unit test for this

PR #457, opened 2018-04-24
 156952821 Fix file input tooltip visibility issue in pre-notice form

K1ll3rF0x:


K1ll3rF0x:
#156952821 Waiting verification from Chrome on macOS users.

PR #462, opened 2018-04-25
156871486 delete pre notice file

mmarkuss:
# Fixed
* Pre-notice attachment file deletion.


ote/src/clj/ote/services/pre_notices/attachments.clj
```diff
@@ -49,6 +49,10 @@
               (fn [out]
                 (io/copy (:input-stream s3-file) out)))})))
 
+(defn delete-from-aws [db {bucket :bucket :as config} attachments]
```
K1ll3rF0x: Maybe delete-from-s3 would be more specific?

ote/src/clj/ote/services/pre_notices/attachments.clj
```diff
@@ -49,6 +49,10 @@
...
+  (doseq [{id ::transit/id file-name ::transit/attachment-file-name}  attachments]
+    (s3/delete-object bucket (str id "_" file-name))))
```
K1ll3rF0x: Better to create a function for generating bucket file keys. This is already used in "upload attachment" "download attachment" and "delete-from-aws" If the filekey format changes at some point, we'd have to remember to change it for all those places.

ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -43,29 +44,43 @@
...
-                           (op/and
+(defn- list-attachments-in-db [db user]
+  (specql/fetch db ::transit/pre-notice-attachment
+                (specql/columns ::transit/pre-notice-attachment)
+                {::modification/created-by (authorization/user-id user)}))
+
+(defn- delete-attachments [db config attachments]
+  (doseq [{id ::transit/id} attachments]
+    (specql/delete! db ::transit/pre-notice-attachment {::transit/id id}))
+  (attachment/delete-from-aws db config attachments))
```
K1ll3rF0x: What happens if the deletion from s3 bucket fails for some reason? 
Pre-notice-attachments are already removed at this point, so we cannot retry the s3 delete query in case of failure. However, I don't think it is a  big issue if there are some orphaned files in the bucket. They can be addressed with s3 lifecycle management.

ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -43,29 +44,43 @@
...
+                       attachments-from-db)]
+    (authorization/with-transport-operator-check
+      db user (::t-operator/id notice)
+      (fn []
+        (tx/with-transaction db
+          (let [form-attachments (:attachments notice)
+                n (-> notice
+                      (dissoc :attachments)
+                      (dissoc ::transit/attachments)
+                      (modification/with-modification-fields ::transit/id user))
```
K1ll3rF0x: "with-sent-field" missing. Have you merged the newest master?

ote/src/clj/ote/services/pre_notices/attachments.clj
mmarkuss: Fixed. Renamed.

ote/src/clj/ote/services/pre_notices/attachments.clj
mmarkuss: Fixed. Key-file function created for common use.

ote/src/clj/ote/services/pre_notices/operator.clj
mmarkuss: S3 call is inside the transaction and failure should rollback the transaction.

ote/src/clj/ote/services/pre_notices/operator.clj
```diff
@@ -43,35 +44,49 @@
...
-               saved-notice (specql/upsert! db ::transit/pre-notice n)
-               notice-id (::transit/id saved-notice)]
-           (log/debug "Save notice: " saved-notice)
-           (specql/update! db ::transit/pre-notice-attachment
-                           {::transit/pre-notice-id notice-id}
-                           (op/and
+(defn save-pre-notice [db user notice config]
+  (let [attachments-from-db (list-attachments-in-db db user)
+        attachments-ids-from-client (set (map ::transit/id (:attachments notice)))
+        to-be-removed (filter #(not (contains? attachments-ids-from-client (::transit/id %)))
```
tatut: won't this remove all attachments in previous notices by the same user?
`list-attachments-in-db` should only return attachments that *are not* linked yet to a notice?

mmarkuss: Fixed to list attachment files which are not linked to any pre-notice or files which are linked to current open pre notice.

PR #469, opened 2018-04-26
157118940 Show pre notice route

tatut:
# Added
- Show pre notice by direct link

### DIFF
--- a/ote/src/cljs/ote/app/controller/pre_notices.cljs
+++ b/ote/src/cljs/ote/app/controller/pre_notices.cljs
@@ -25,6 +25,8 @@
               :on-failure (tuck/send-async! ->ServerError)})
   :loading)
 
+(declare ->ShowPreNotice)
+
 (tuck/define-event LoadAuthorityPreNotices []
   {:path [:pre-notices]}
   (comm/get! "pre-notices/authority-list"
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/app/controller/pre_notices.cljs
```diff
@@ -36,8 +38,10 @@
 (defmethod routes/on-navigate-event :pre-notices [_]
   (->LoadOrganizationPreNotices))
 
-(defmethod routes/on-navigate-event :authority-pre-notices [_]
-  (->LoadAuthorityPreNotices))
+(defmethod routes/on-navigate-event :authority-pre-notices [{params :params}]
```
rckopteri: I think that this is a bit scary. Not because given parameters arent mandatatory but because it call entirely different event. And by event I mean "page". So for me this would be easier to understand if we had
route event for "authority-pre-notices" and "auhtority-pre-notice/:id" (which is not restful naming convention).

ote/src/cljs/ote/app/routes.cljs
```diff
@@ -64,7 +65,15 @@
                (if event
                  (binding [tuck/*current-send-function*
                            #(swap! state/app (flip tuck/process-event) %)]
-                   (tuck/process-event event app))
+                   (if (vector? event)
```
rckopteri: When this is used?

PR #473, opened 2018-04-27
Change the test to use stops with larger distance to prevent the stops covering each other.

mmarkuss:
# Fixed
* Sea route stop are overlapped easily which may cause the test fail. Changed the test to use stops outside Finland with larger distances.

### DIFF
--- a/cypress/integration/route_spec.js
+++ b/cypress/integration/route_spec.js
@@ -30,10 +30,15 @@ describe('Sea route tests', function () {
     });
 
     it('Add stops', function () {
-        cy.get('img[title="OULUN SATAMA: Vihreäsaari"]').click();
-        cy.get('img[title="RAAHE: Rautaruukki"]').click();
-        cy.get('img[title="KALAJOKI: Itäkenttä"]').click();
-        cy.get('img[title="VAASA: Lassenlaituri"]').click();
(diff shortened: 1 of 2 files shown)


cypress/integration/route_spec.js
```diff
@@ -30,10 +30,15 @@ describe('Sea route tests', function () {
...
-        cy.get('img[title="OULUN SATAMA: Vihreäsaari"]').click();
-        cy.get('img[title="RAAHE: Rautaruukki"]').click();
-        cy.get('img[title="KALAJOKI: Itäkenttä"]').click();
-        cy.get('img[title="VAASA: Lassenlaituri"]').click();
+        cy.get('.leaflet-control-zoom-out').click().wait(200);
+        cy.get('.leaflet-control-zoom-out').click().wait(200);
+        cy.get('.leaflet-control-zoom-out').click().wait(200);
+        cy.get('.leaflet-control-zoom-out').click().wait(200);
+        cy.get('.leaflet-control-zoom-out').click();
+        cy.get('img[title="Puerto Mont"]').click();
```
rckopteri: isn't it Puerto montt, with two tt

mmarkuss: It is indeed. Good catch! :)

PR #475, opened 2018-05-02
157247809 Fix scheduled task time bug

tatut:
# Fixed
- 1st scheduled task time was skipped if restarting before it

### DIFF
--- a/ote/src/clj/ote/tasks/pre_notices.clj
+++ b/ote/src/clj/ote/tasks/pre_notices.clj
@@ -6,7 +6,6 @@
             [clj-time.core :as t]
             [clj-time.format :as format]
             [clj-time.coerce :as coerce]
-            [clj-time.periodic :refer [periodic-seq]]
             [clojure.string :as str]
             [hiccup.core :refer [html]]
             [ote.db.transit :as transit]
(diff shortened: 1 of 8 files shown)


ote/src/clj/ote/tasks/pre_notices.clj
```diff
@@ -130,22 +130,18 @@
              (log/warn "Error while sending a notification" e))))))))
 
 
-(defrecord PreNoticesTasks [at email-config interval]
+(defrecord PreNoticesTasks [email-config]
   component/Lifecycle
   (start [{db :db :as this}]
     (assoc this
-      ::stop-tasks [(chime-at (drop 1 (periodic-seq at (t/seconds
-                                                         (hours->seconds (or interval
```
K1ll3rF0x: Why did you remove the usage of interval option? It is much easier to change task intervals from config.edn than in code, no new build is needed.

tatut: I think we don't need it, it is even easier to call the timed function from the RPEL for testing purposes.

The `daily-at` gives a sequence of daily times (without an interval).

K1ll3rF0x: Ok, could you also remove the interval settings from config.end and aws config.edn templates?

PR #476, opened 2018-05-03
157110973 pre notice attachment button

mmarkuss:
# Changed
* Replace add and delete buttons with combined functionality. After adding file 'Add' button changes to 'Delete' button.


ote/src/cljs/ote/app/controller/pre_notices.cljs
```diff
@@ -252,6 +252,11 @@
                                            :on-failure (tuck/send-async! ->ServerError)})
   app)
 
+(define-event DeleteAttachment [row]
```
rckopteri: Could you rename row -> row-index

PR #492, opened 2018-05-08
157311328 Improved GTFS import

tatut:
# Changed
- GTFS import now saves all the data

### DIFF
--- a/database/src/main/resources/db/migration/V1_82__gtfs.sql
+++ b/database/src/main/resources/db/migration/V1_82__gtfs.sql
@@ -1,0 +1,40 @@
+ALTER TABLE "gtfs-stop"
+  ADD COLUMN "wheelchair-boarding" INTEGER CHECK ("wheelchair-boarding" BETWEEN 0 AND 2);
+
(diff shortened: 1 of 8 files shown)


database/src/main/resources/db/migration/V1_82__gtfs.sql
```diff
@@ -0,0 +1,40 @@
...
+  "departure-time"    INTERVAL,
+  "stop-sequence"     INTEGER,
+  "stop-headsign"     TEXT,
+  "pickup-type"       INTEGER,
+  "drop-off-type"     INTEGER,
+  "shape-dist-traveled" NUMERIC
+);
+
+CREATE TYPE "gtfs-trip-info" AS (
+  "trip-id"               TEXT,
```
rckopteri: This links gtfs_trip to this type? gtfs_trip.id is int type. This should be too.

K1ll3rF0x: trip_id can contain characters.

PR #496, opened 2018-05-14
156871396 chip

rckopteri:
# Added
* Take chip-input into use in service search sub-type

### DIFF
--- a/ote/src/cljs/ote/views/service_search.cljs
+++ b/ote/src/cljs/ote/views/service_search.cljs
@@ -258,12 +258,12 @@
         sub-types-to-list (fn [data]
                             (keep (fn [val]
                                     (let [subtype (:sub-type val)]
-                                      (when (not= :other subtype)
+                                      (when-not (= :other subtype)
                                         (into (sorted-map)
                                               (-> val
                                                   (dissoc :sub-type)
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -254,7 +254,17 @@
                         operators :operators
                         :as service-search}]
   (let [sub-type (tr-key [:enums ::t-service/sub-type]
-                         [:enums ::t-service/type])]
+                         [:enums ::t-service/type])
+        sub-types-to-list (fn [data]
+                            (keep (fn [val]
+                                    (let [subtype (:sub-type val)]
+                                      (when (not= :other subtype)
```
K1ll3rF0x: Using when-not would be prettier.

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -254,7 +254,17 @@
...
+                                        (into (sorted-map)
+                                              (-> val
+                                                  (dissoc :sub-type)
+                                                  (assoc :value subtype)
```
K1ll3rF0x: You could combine these assocs into one assoc invocation.

PR #497, opened 2018-05-14
157274612 Add per date hashes to GTFS package import

tatut:
# Added
- GTFS import: the "shape" of traffic per day is calculated and stored as a hash

# Datamodel
- Added "gtfs-date-hash" table which stores per day hash

### DIFF
--- a/ote/src/clj/ote/integration/import/gtfs.clj
+++ b/ote/src/clj/ote/integration/import/gtfs.clj
@@ -155,11 +155,11 @@
 
 ;; PENDING: this is for local testing, truncates *ALL* GTFS data from the database
 ;;          and reads in a local GTFS zip file
-(defn test-hsl-gtfs []
+#_(defn test-hsl-gtfs []
   (let [db (:db ote.main/ote)]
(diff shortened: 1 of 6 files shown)


ote/src/clj/ote/integration/import/gtfs.clj
```diff
@@ -155,7 +159,7 @@
   (let [db (:db ote.main/ote)]
     (clojure.java.jdbc/execute! db ["TRUNCATE TABLE gtfs_package RESTART IDENTITY CASCADE"])
     (clojure.java.jdbc/execute! db ["INSERT INTO gtfs_package (id) VALUES (1)"])
-    (let [bytes (with-open [in (io/input-stream "hsl_gtfs.zip")]
+    (let [bytes (with-open [in (io/input-stream "/Users/tatuta/Downloads/gtfs_tampere (2).zip" #_"hsl_gtfs.zip")]
```
rckopteri: Not sure but I think that our ec2 instance has different path for zip files.

tatut: This is a function for REPL testing only and is commented out,

PR #516, opened 2018-05-22
157687729 Capitalize operation-area in search result page

rckopteri:
# Changed
* Capitalize operation area texts in search page

### DIFF
--- a/ote/src/cljs/ote/views/service_search.cljs
+++ b/ote/src/cljs/ote/views/service_search.cljs
@@ -345,6 +345,7 @@
           :open-on-focus? true})]
       filters]]))
 
+
 (defn service-search [e! app]
   (e! (ss/->InitServiceSearch))
   (fn [e! {{results :results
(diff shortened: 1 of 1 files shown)


ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -241,13 +241,23 @@
 
       [operator-result-chips e! chip-results]]))
 
+(defn- capitalize-operation-area [sentence]
+  (str/replace sentence #"([^A-Öa-ä0-9_])(\w)"
```
tatut: why are capital letters in a different range than lower case letters? what about Å?

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -241,13 +241,23 @@
 
       [operator-result-chips e! chip-results]]))
 
+(defn- capitalize-operation-area [sentence]
```
K1ll3rF0x: This function works only for operation areas that is formed using multiple words and numbers such as postal codes.
E.g. "90570 oulu" would work, but "helsinki" would not.

PR #517, opened 2018-05-22
157547079 Load region geojson when editing pre-notice

tatut:
# Fixed
- Region geometry is loaded when editing a pre-notice

### DIFF
--- a/ote/src/cljs/ote/app/routes.cljs
+++ b/ote/src/cljs/ote/app/routes.cljs
@@ -47,6 +47,17 @@
 
 (defmethod on-navigate-event :default [_] nil)
 
+(defn- send-startup-events [event]
+  (let [e! (fn e! [event]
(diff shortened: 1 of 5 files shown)


ote/src/cljs/ote/app/routes.cljs
```diff
@@ -47,6 +47,17 @@
 
 (defmethod on-navigate-event :default [_] nil)
 
+(defn- send-startup-events [event]
+  (let [e! (fn e! [event]
+             (binding [tuck/*current-send-function* e!]
```
K1ll3rF0x: What is happening here?

tatut: Tuck relies on a dynamic binding for *current-send-function* to be present (because the `e!` is not passed in to process event)

We need to bind it to the same function when processing so that it is available.

PR #518, opened 2018-05-22
Update country codes to correct ISO 3166-1 form.

mmarkuss:
# Fixed
* Update country codes to correct ISO 3166-1 form.


database/src/main/resources/db/migration/V1_87__country_code_fix.sql
```diff
@@ -0,0 +1,115 @@
+-- Update country codes to correct ISO 3166-1 form.
+-- Remove duplicate Indian Ocean Territories
+-- Update Yhdistynyt Kuningaskunta to Iso-Britannia
+
+delete country where nameeng = 'Indian Ocean Territories';
+
+update country set code = 'AE' where nameeng = 'United Arab Emirates';
+update country set code = 'AG' where nameeng = 'Antigua and Barbuda';
+update country set code = 'AG' where nameeng = 'Antigua and Barbuda';
```
K1ll3rF0x: Why there are some update lines twice? For example this line here. Below there are more examples.

mmarkuss: Good catch. Duplicates removed.

PR #530, opened 2018-05-29
157654432 Prevent gtfs updates and downloads for listed operators

mmarkuss:
# Added
- Prevent gtfs updates and downloads for listed operators

### DIFF
--- a/aws/ansible/environments/production/group_vars/ote/vars
+++ b/aws/ansible/environments/production/group_vars/ote/vars
@@ -11,4 +11,4 @@ email_from: noreply@finap.fi
 email_smtp_server: email-smtp.eu-west-1.amazonaws.com
 email_user: "{{vault_email_user}}"
 email_password: "{{vault_email_password}}"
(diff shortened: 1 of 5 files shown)


aws/ansible/environments/production/group_vars/ote/vars
```diff
@@ -10,4 +10,5 @@ pre_notice_attachment_bucket: "napote-pre-notice-attachment-prod"
 email_from: noreply@finap.fi
 email_smtp_server: email-smtp.eu-west-1.amazonaws.com
 email_user: "{{vault_email_user}}"
-email_password: "{{vault_email_password}}"
\ No newline at end of file
+email_password: "{{vault_email_password}}"
+no_gtfs_update_for_operators: "#{}"
```
rckopteri: Should we add hsl and tampere in this first commit?

mmarkuss: I thought we add those afterwards but it is possible now also.

PR #545, opened 2018-06-04
157994008 Pre-notice attachment file content type validation

K1ll3rF0x:
# Added
* File content validation for pre-notice form attachments. Allowed file types are now pdf, jpeg and png.

# Environment
* Increased nginx max request body size to 33MB.

### DIFF
--- a/ote/src/clj/ote/services/pre_notices/attachments.clj
+++ b/ote/src/clj/ote/services/pre_notices/attachments.clj
@@ -46,7 +46,7 @@
         (case msg
           "Invalid file type"
           {:status 422
-           :body (str "Invalid file type.")}
+           :body "Invalid file type."}
           {:status 500
            :body msg})))))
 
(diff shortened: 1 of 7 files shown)


ote/src/clj/ote/services/pre_notices/attachments.clj
```diff
@@ -11,24 +11,44 @@
...
+
+(def allowed-mime-types ["application/pdf" "image/jpeg" "image/png"])
 
 (defn- generate-file-key [id filename]
   (str id "_" filename))
 
+(defn validate-file [{:keys [tempfile]}]
+  (let [path (.toPath tempfile)
+        content-mime (Files/probeContentType path)]
+    (when-not (some #(= % content-mime) allowed-mime-types)
```
tatut: If you changed `allowed-mime-types` from a vector to a set, you could simply check `(allowed-mime-types content-mime)`

tatut: We should check (by name, I guess) on the frontend also so we don't upload something that is then rejected.

K1ll3rF0x: Yes, true that. And a naive file-extension based validation could be a good choise for the front-end.

PR #551, opened 2018-06-06
157929105 Remove service publishment and deletion from CKAN

mmarkuss:
# Changed
* Remove service publishment and deletion from CKAN


mmarkuss:
Should we also remove ckan-dataset-id and ckan-resource-id columns from transport-service table?

PR #563, opened 2018-06-12
157654432 Operator blacklist for external interfaces

mmarkuss:
# Fixed
* Operator blacklist for external interfaces

### DIFF
--- a/ote/src/clj/ote/tasks/gtfs.sql
+++ b/ote/src/clj/ote/tasks/gtfs.sql
@@ -1,14 +1,15 @@
 -- name: select-gtfs-urls-update
 SELECT eid.id as id, (eid."external-interface").url, eid.format[1], eid.license,
        ts."transport-operator-id" as "operator-id", ts.id as "ts-id", eid."gtfs-imported" as "last-import-date"
-  FROM "transport-service"  ts, "external-interface-description"  eid
+  FROM "external-interface-description"  eid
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/tasks/gtfs.sql
```diff
@@ -1,14 +1,15 @@
 -- name: select-gtfs-urls-update
 SELECT eid.id as id, (eid."external-interface").url, eid.format[1], eid.license,
        ts."transport-operator-id" as "operator-id", ts.id as "ts-id", eid."gtfs-imported" as "last-import-date"
-  FROM "transport-service"  ts, "external-interface-description"  eid
+  FROM "external-interface-description"  eid
+  JOIN "transport-service" ts ON eid."transport-service-id" = ts.id
+  JOIN "transport-operator" top ON  top.id = ts."transport-operator-id"
  WHERE ts."published?" = TRUE
-   AND ts.id = eid."transport-service-id"
+   AND top.id NOT IN (:blacklist)
```
rckopteri: Could this work:
(:blacklist::TEXT IS NOT NULL OR top.id NOT IN (:blacklist))
Then you could set variable :blacklist to nil

tatut: :blacklist is not text (it is a collection parameter), so I don't think that will work

PR #565, opened 2018-06-14
157910823 edit user

rckopteri:
# Added
* Edit user dialog to admin panel
* Delete user dialog to admin panel


ote/src/cljs/ote/views/admin/users.cljs
```diff
@@ -0,0 +1,139 @@
...
+            [ote.ui.common :refer [linkify]]
+            [ote.time :as time]
+            [cljs-react-material-ui.icons :as ic]
+            [reagent.core :as r]
+            [ote.ui.common :as ui-common]
+            [ote.ui.common :as common-ui]))
+
+(defn- delete-user-action [e! {:keys [id show-delete-modal?] :as user}]
+  [:span
+   (.log js/console " delete-user-action user:  " (pr-str user))
```
mmarkuss: Console logging could be removed.

mmarkuss: Seems to be removed in later task.

ote/src/clj/ote/nap/users.sql
```diff
@@ -59,3 +59,7 @@ SELECT group_id
   FROM group_extra
  WHERE key = 'transit-authority?'
    AND value = 'true';
+
+-- name: delete-user!
+-- Delete user from the database
+DELETE FROM "user" as u WHERE u.id = :id;
```
tatut: What happens if user is in groups or has authored services? Do they need to be deleted first?

rckopteri: Yes. This only deletes user. All other data must be deleted separately.
I thought that it would be bad if operator had two users and one of them was removed and all the data from the operator would have been gone.

PR #621, opened 2018-07-16
159064940 Handle GTFS interface URL errors and store the errors messages.

K1ll3rF0x:
# Changed
* We'll now check and store also GTFS external interface URL errors.


K1ll3rF0x:
Adding also kalkati lambda error catching for admin panel.

PR #653, opened 2018-08-22
159885008 Fix critical bugs in timestamp formatting in ote.time.

K1ll3rF0x:
# Fixed
* Fixed bugs in timestamp conversion in ote.time. Previously, the convert functions just returned todays date no matter what timestring was used as a parameter.

### DIFF
--- a/ote/src/cljc/ote/time.cljc
+++ b/ote/src/cljc/ote/time.cljc
@@ -31,26 +31,29 @@
 (s/def ::specql-data-types/time #(instance? Time %))
 
 #?(:cljs
-(defn format-timestamp-for-ui [time]
-  (if  (nil? time)
+(defn format-timestamp-for-ui [timestring]
+  (if  (nil? timestring)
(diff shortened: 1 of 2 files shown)


ote/src/cljc/ote/time.cljc
```diff
@@ -31,28 +31,37 @@
...
-(defn format-timestamp-for-ui [time]
-  (if  (nil? time)
-    " " ;: if nil - print empty string
-    (->> time
+   (defn format-timestamp-for-ui [timestamp]
+     (if (nil? timestamp)
+       " " ;: if nil - print empty string
+       (->>
+         (if (string? timestamp)
+           (coerce/from-string timestamp)
```
tatut: This function should NOT be called with a string parameter. I think we should fix erroneus call sites instead.

K1ll3rF0x: Yes, I can do that. Will change the original "timestamp" parameter name too, because can be misleading.

PR #676, opened 2018-09-24
Add 200ms sleep after every pre-notice email

rckopteri:
# Added
* Add sleep to email sending to prevent SES sending limits to be exeeded


tatut:
good to go 👌

PR #691, opened 2018-10-02
Make sure that docker container ports are exposed only on localhost.

K1ll3rF0x:
# Fixed
* Fixed security issue with docker-compose. By default docker exposes container ports to all interfaces. We do not want that, so now they are strictly locahost only. https://www.techrepublic.com/article/how-to-fix-the-docker-and-ufw-security-flaw/

### DIFF
--- a/nap/docker/docker-compose.yml
+++ b/nap/docker/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       - napote-solr
       - napote-redis
     ports:
-      - "${CKAN_PORT}:5000"
+      - "127.0.0.1:${CKAN_PORT}:5000"
     environment:
       - CKAN_SQLALCHEMY_URL=postgresql://ckan:@napotedb:5432/napote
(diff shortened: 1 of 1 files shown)


K1ll3rF0x:
Seems that there are still some problems that need to be resolved before merging this change. These errors appear after trying to start lein repl when using the new docker-compose:
`"HikariDataSource (HikariPool-1)"], :config {:url "jdbc:postgresql://localhost:5432/napote", :username "napote", :password ""}}}, :system #<SystemMap>}
	at clojure.core$ex_info.invokeStatic(core.clj:4725)
....
Caused by: java.net.BindException: Address already in use
	at sun.nio.ch.Net.bind0(Native Method)
	....
No nREPL ack received`

K1ll3rF0x:
Ok, seems that the issue was only on my machine. It was caused by another repl process that was running in the background for some reason. :man_facepalming:

PR #696, opened 2018-10-08
160509513 show first stop list by default

rckopteri:
# Added
* Possibility to add svg icons when needed icon is on in Material-ui-icon font
* Select first trip by default when route is selected in gtfs change visualization page
* Own loading? parameter to all async server request. Page is rendered when all loading? events are false.
   
# Fixed
* Select table row didn't work correctly when user clicked off the table. Fixed by adding a nil check to row-select event.
   
# Changed
* Update H1 and H2 elements


ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -694,11 +694,22 @@
...
+                                        route-trips-for-date2-loading?
+                                        route-calendar-hash-loading? route-differences-loading? routes-for-dates-loading?
+                                        service-changes-for-dates-loading?
+                                        highlight service-info
                                         changes selected-route compare open-sections]
-                                 :as transit-visualization}]
+                                 :as   transit-visualization}]
   [:div
-   (when (not loading?)
+   (when (and
```
tatut: Why does to UI need to know about every loading state? I think it would be cleaner if the events just set one boolean for the UI to check if it is "ready"

ote/src/cljs/ote/ui/table.cljs
```diff
@@ -28,8 +28,8 @@
                }
               (when on-select
                 {:on-row-selection (fn [selected-rows]
-                                     (let [rows (vec rows)]
-                                       (on-select (map (partial nth rows) selected-rows))))})
+                                     (when (not (empty? selected-rows)) (let [rows (vec rows)]
```
tatut: Use `seq` instead of `(not (empty? ...))` format the let to start on its own line

ote/src/cljs/ote/ui/page.cljs
```diff
@@ -20,5 +20,5 @@
    [:div.container ; (stylefy/use-style style/content)
     top-link
     [:h1 title]
-    [:div.page-controls-content
+    [:div.page-controls-content {:style {:padding-bottom "1.25rem"}}
```
tatut: Use stylefy

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -280,6 +312,7 @@
                   (select-keys route #{:gtfs/added-trips :gtfs/removed-trips
                                        :gtfs/trip-stop-sequence-changes
                                        :gtfs/trip-stop-time-changes}))
+        ;TODO: Dissoc selected route
```
tatut: Fix todo

ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -723,8 +734,8 @@
 
          (when (and hash->color date->hash)
            [:span
+            ;; Select first trip to show stops
```
tatut: I don't understand the above comment, what does it reference... should it be in a controller event?

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -17,6 +17,19 @@
    "#ec8fb5" "#23dbe1" "#a4515b" "#169294" "#fd5925" "#3d4e92" "#f4d403"
    "#66a1e5" "#d07d09" "#9382e9" "#b9cf84" "#544437" "#f2cdb9"])
 
+(defn loaded-from-server? [{:keys [route-lines-for-date-loading? route-trips-for-date1-loading?
+                                   route-trips-for-date2-loading? route-calendar-hash-loading?
+                                   route-differences-loading? routes-for-dates-loading?
+                                   service-changes-for-dates-loading?]
+                            :as   transit-visualization}]
+  (and (not route-lines-for-date-loading?)
+  (not route-trips-for-date1-loading?)
```
tatut: all forms in `(and ...)` should be indented at same level

PR #714, opened 2018-10-16
161173800 Add missing COALESCEs in fetch-route-differences.

K1ll3rF0x:
# Fixed
* Added missing COALESCEs in fetch-route-differences. Fixes missing changes under calendar when manually choosing dates for comparison.


K1ll3rF0x:
Leaving this as WIP for now, encountered other problems while fixing this one.

PR #732, opened 2018-10-31
Skip Static holidays

rckopteri:
# Added
* List of static holidays
* Skip static holidays in automatic gtfs data parsing
* Test to test that skip holidays works


ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -386,6 +386,15 @@
 
                      :gtfs/package-ids package-ids})))
 
+(defn override-static-holidays [date-route-hashes]
+  (map (fn [{:keys [date] :as row}]
+         (if-let [holiday-id (transit-changes/is-holiday? date)]
+           (do
+             (println "changed holiday hash " (:hash row) " => " holiday-id " for " date)
```
tatut: remove `println` and surrounding `do`

PR #738, opened 2018-11-08
 Geotools osgeo repository uri to https

philosotractor:
# Changed
* Change repository uri to "https:" because new figwheel requires TLS.
Otherwise lein figwheel fails.


tatut:
How does figwheel require https? Osgeo is not used on the frontend.

philosotractor:
Commit and PR renamed

PR #739, opened 2018-11-09
Fix open figwheel websocket in dev setup

erno:


philosotractor:
LGTM, though the two commits could  be squashed :)

PR #751, opened 2018-11-16
161439932 remove old changes from detection email

rckopteri:
# Fixed
* Remove old changes from detected changes email. chg.date in with clau…se will remove most of the packages out of "inner table" and it will speed up the query.
* By taking gtfs_package into account we can check that the relevant new changes happens for packets that are downloaded (created) in current day.

### DIFF
--- a/ote/src/clj/ote/tasks/pre_notices.sql
+++ b/ote/src/clj/ote/tasks/pre_notices.sql
@@ -24,6 +24,8 @@ WITH changes_with_regions AS (
                     FROM gtfs_package p
                    WHERE p.id = ANY(chg."package-ids")) x) AS "finnish-regions"
     FROM "gtfs-transit-changes" chg
+    WHERE chg.date = CURRENT_DATE
+
 )
 SELECT to_char(chg."change-date", 'dd.mm.yyyy') as "change-date",
        ("change-date" - CURRENT_DATE) AS "days-until-change",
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/tasks/pre_notices.sql
```diff
@@ -37,8 +39,11 @@ SELECT to_char(chg."change-date", 'dd.mm.yyyy') as "change-date",
        ts.id AS "transport-service-id"
   FROM changes_with_regions chg
   JOIN "transport-service" ts ON ts.id = chg."transport-service-id"
-  JOIN "transport-operator" op ON op.id = ts."transport-operator-id"
+  JOIN "transport-operator" op ON op.id = ts."transport-operator-id",
+       gtfs_package p
  WHERE chg.date = CURRENT_DATE
+    AND p.id = ANY(chg."package-ids")
```
philosotractor: Indentation :)

ote/src/clj/ote/tasks/pre_notices.sql
```diff
@@ -37,8 +39,11 @@ SELECT to_char(chg."change-date", 'dd.mm.yyyy') as "change-date",
...
+   AND p.created > CURRENT_DATE - interval '8 hours'
```
philosotractor: Will this assume package creation done at fixed time 08.15 and break if that changes? 
If so, instead of shifting 8h could this use CURRENT_DATE anbd time of hour 00.00 or 00.01?

rckopteri: I'm not sure, but I think that we have a mismatch in datetime values in db. Some of them are created using now() function in db and some of them using java.datetime. And those differences might cause mismatches  when comparing dates in database.

Our automatic change detection is timed to start at 05.15. And gtfs import is timed to start at 00.00,  But due to my assumption that times are in different timezones I add 8 hour window to skip any timezone differences (3 hours at summer, 2 hours at winter time)

ote/test/clj/ote/tasks/pre_notices_test.clj
```diff
@@ -71,16 +71,27 @@
...
                     {:gtfs/transport-service-id 1
-                     :gtfs/date (java.util.Date.)
-                     :gtfs/current-week-date (java.util.Date.)
-                     :gtfs/different-week-date (tc/to-sql-date (time/days-from (time/now) 70))
-                     :gtfs/change-date (tc/to-sql-date (time/days-from (time/now) 67))
-                     :gtfs/package-ids [1]
-                     :gtfs/removed-routes 1
-                     :gtfs/added-routes 2
-                     :gtfs/changed-routes 3})
+                     :gtfs/date                 (java.util.Date.)
```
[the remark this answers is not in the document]
rckopteri: changed

ote/src/clj/ote/tasks/pre_notices.sql
rckopteri: fixed

PR #755, opened 2018-11-21
162012032 icon changes to transit vis. view / palvelukohtaisten muutosten näkymän muutoksia

erno:
v2 of the icon change, specs on the pivotal ticket.


erno:
I didn't yet figure out how to generate test data showing changes in my dev setup, so I haven't checked whether the red/green colors are still working on the change rows.

rckopteri:
gtfs-transit-changes table is responsible for showing those changes. We can check it out next time.

PR #757, opened 2018-11-27
162234002 Calculate route-hash-id for all routes

rckopteri:
# Added
* Add postgres functions to calculate route-hash-id for all routes using default "short-long-headsign" pattern.


database/src/main/resources/db/migration/V1_122__add_route_hash_ids_to_all_packages.sql
```diff
@@ -0,0 +1,20 @@
+-- Create function that generates routes to detection-route table using default route-hash-id (short-long-headsign)
+CREATE OR REPLACE FUNCTION generate_route_hash_ids_for_packages() RETURNS VOID AS $$
+DECLARE
+ p_id RECORD;
+BEGIN
+
+  -- Delete all
+  DELETE FROM "detection-route";
```
philosotractor: Not needed? :)

rckopteri: Fixed

philosotractor:
LGTM

PR #759, opened 2018-11-30
Improve admin panel ui and upload gtfs package for service

rckopteri:
# Added
* Upload gtfs package for service at a given day
* Improve admin panel ui
* List all services with route-hash-id


ote/src/cljs/ote/app/controller/admin.cljs
```diff
@@ -474,6 +483,41 @@
              {:on-success #(.log js/console %)})
   app)
 
+(define-event UploadResponse [response]
+  {}
+  (assoc app :flash-message "Paketti ladattu"))
```
philosotractor: Localistation string?

PR #762, opened 2018-12-04
162383037 route hash id options

rckopteri:
# Added
* Few more options to calculate route-hash-id
   
# Fixed
* Null pointer when we can't find trips for route
   
# Changed
* route long name, route short name and trip headsign queries to use only route-hash-id as a key

### DIFF
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -43,16 +43,9 @@
       "short-long-headsign")))
 
 (defn use-old-route-key [type]
-  (if (or (= type "short-long") (= type "route-id"))
-    false
-    true))
-
-(defn route-key
-  "Route key is a vector of [short-name long-name headsign]"
-  [type]
-  (if (use-old-route-key type)
-    (juxt :route-short-name :route-long-name :trip-headsign)
-    :route-hash-id))
+  (if (= type "short-long-headsign")
+    true
+    false))
 
 (defn calculate-route-hash-id-for-service
   "We support only few different hash calculation types. [short-long-headsign, short-long, route-id]"
@@ -62,7 +55,7 @@
     (specql/delete! db :gtfs/detection-service-route-type {:gtfs/transport-service-id service-id})
     (specql/insert! db :gtfs/detection-service-route-type
                     {:gtfs/transport-service-id service-id
-                     :gtfs/route-hash-id-type type})
+                     :gtfs/route-hash-id-type   type})
     (doall
       ;; Calculate route-hash-id:s again to detection-route table for given service.
       (for [package-id package-ids]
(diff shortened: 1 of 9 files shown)


ote/src/cljs/ote/views/admin/detected_changes.cljs
```diff
@@ -141,19 +141,15 @@
...
   [:div
     [:h4 "Lataa palvelulle gtfs tiedosto tietylle päivälle"]
     [form/form
      {:update!   #(e! (admin-controller/->UpdateUploadValues %))
       :footer-fn (fn [data]
-                   [:span
-                    #_ [ui/raised-button {:primary  true
-                                       :on-click #(e! (admin-controller/->ForceRouteHashCalculationForService))
-                                       :label    "Käynnistä gtfs lataus"}]])}
+                   [:span])}
```
philosotractor: If footer is not needed, you could just remove :footer-fn. That way there's no redundat dom elements.

ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -25,7 +25,10 @@
 (defn merge-by-closest-time [time-fn left-items right-items]
   (let [left-items-with-closest (mapv #(item-with-closest-time time-fn % right-items) left-items)
         right-items-with-closest (mapv (comp flip-vec #(item-with-closest-time time-fn % left-items)) right-items)
-        time-diff (fn [[l r]] (time/time-difference (time-fn l) (time-fn r)))
+        time-diff (fn [[l r]]
+                    (if (and (nil? (complement r)) (nil? (complement l)))
```
philosotractor: ```complement```here doesn't do what was expected, should return always ```false```

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -62,7 +55,7 @@
     (specql/delete! db :gtfs/detection-service-route-type {:gtfs/transport-service-id service-id})
     (specql/insert! db :gtfs/detection-service-route-type
                     {:gtfs/transport-service-id service-id
-                     :gtfs/route-hash-id-type type})
+                     :gtfs/route-hash-id-type   type})
```
philosotractor: Nitpicking, but the extra space could be removed

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -428,9 +423,9 @@
 
 (defn- debug-print-change-stats [all-routes route-changes type]
   (doseq [r all-routes
-          :let [key ((route-key type) r)
+          :let [key (:route-hash-id r)
                 {:keys [changes no-traffic-start-date no-traffic-end-date]
-                 :as route} (route-changes key)]]
+                 :as   route} (route-changes key)]]
```
philosotractor: Extra space before `route`

PR #767, opened 2018-12-12
162080404 improve route detection

rckopteri:
# Added
* Route-hash-id as a default way to match routes between gtfs packages

# Fixed
* "zero" row detection by ensuring that routes are matched between packages and date hashes are ordered similar way between packages using external-interface-description-id as a ordering key

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -333,6 +333,9 @@
   (let [service-id (get-in app [:params :service-id])
         current-week-date (or (get-in app [:transit-visualization :changes :gtfs/current-week-date])
                               (t/now))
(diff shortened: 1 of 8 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -345,8 +348,7 @@
                             {:long-name long})
                           (when-let [headsign (:gtfs/trip-headsign route)]
                             {:headsign headsign})
-                          (when-let [route-hash-id (:gtfs/route-hash-id route)]
-                            {:route-hash-id route-hash-id}))
+                            {:route-hash-id route-hash-id})
```
philosotractor: Minor code layout issue for `{:route-hash-id route-hash-id}`, could be more left & on top

PR #794, opened 2018-12-31
162807267 does company exists

rckopteri:
# Added
* Added check that business-id must be valid and unique before data can be fetched from ytj using given business-id.

### DIFF
--- a/ote/src/cljs/ote/views/transport_operator_ytj.cljs
+++ b/ote/src/cljs/ote/views/transport_operator_ytj.cljs
@@ -61,16 +61,16 @@
        :should-update-check form/always-update
        :on-blur #(e! (to/->EnsureUniqueBusinessId (-> % .-target .-value)))}
 
-      ; show ytj-button only if business-id is not taken and if business-id is valid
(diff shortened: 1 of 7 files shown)


ote/src/cljs/ote/views/transport_operator_ytj.cljs
```diff
@@ -58,15 +58,19 @@
...
-       :label     (tr [:organization-page :fetch-from-ytj])
-       :primary   true
-       :secondary true
-       :on-click  #(e! (to/->FetchYtjOperator (::t-operator/business-id operator)))
-       :disabled  (ytj-loading? state)}
+       :should-update-check form/always-update
+       :on-blur #(e! (to/->EnsureUniqueBusinessId (-> % .-target .-value)))}
+
+      ; show ytj-button only if business-id is not taken and if business-id is valid
+      (when (and (nil? (get-in state [:transport-operator :ote.ui.form/errors ::t-operator/business-id]))
```
philosotractor: Instead of hiding it might be more consistent for UI and UX to simply disable the button.

rckopteri: yep. good idea. I changed this.

PR #796, opened 2019-01-02
162755018 delete operator

rckopteri:
# Changed
* Changed how operator delete works. It is now possible from the checkbox group and for "single" operators delete button is at the bottom of the page.
* I also added a check that operator cannot be removed if it still has services.


ote/src/clj/ote/services/transport.clj
```diff
@@ -111,11 +111,19 @@
...
+                                        #{::t-service/id}
+                                        {:t-service/transport-operator-id id})]
+    ;; delete only if operator-services = nil
+    (if (nil? operator-services)
+      (authorization/with-transport-operator-check
+        db user id
+        #(do
+           (operators/delete-transport-operator db {:operator-group-name (str "transport-operator-" id)})
+           id))
+    {:status 401
```
philosotractor: If we're nitpicking, response should be 403 .
401 means it's an authentication problem and would succeed with proper authentication headers.  403 means request itself was valid and authenticated but server rejected it due to some logic reason

401: https://tools.ietf.org/html/rfc7235#section-3.1
403: https://tools.ietf.org/html/rfc2616#section-10.4.4

ote/src/cljs/ote/app/controller/transport_operator.cljs
```diff
@@ -177,7 +177,18 @@
...
               app)
 
-(define-event ToggleTransportOperatorDeleteDialog []
+(define-event ToggleListTransportOperatorDeleteDialog [operator]
+              {}
+              (update-in app [:transport-operator :transport-operators-to-save]
+                         (fn [coll]
+                           (for [c coll]
+                             (update c :show-delete-dialog?
+                                     #(if (and (or (nil? (:show-delete-dialog? operator)) (= false (:show-delete-dialog? operator)))
```
philosotractor: Shorter form with same result is
```(not(true? (:show-delete-dialog? operator)))```

File: ote/src/cljs/ote/views/transport_operator_ytj.cljs [hunk 3]
```diff
@@ -24,7 +24,14 @@
             [ote.style.base :as style-base]
             [ote.ui.common :as uicommon]))
 
-(defn- delete-operator [e! operator]
+(defn- delete-operator [e! operator service-vector]
+  ;; When operator is passed from :transport-operators-to-save list they do not have business-id
+  (let [toggle-dialog (if (nil? (::t-operator/business-id operator))
```
philosotractor: Instead of resolving toggle callback in delete-operator, could the caller pass the toggle callback as an argumen to this? That way there's no need to resolve it here.

ote/src/clj/ote/services/transport.clj
rckopteri: Agreed. I thought about this, but somehow I decided that 401 close enough :)

ote/src/cljs/ote/ui/form_fields.cljs
```diff
@@ -886,6 +886,54 @@
...
+     (when (not (false? header?))
+       [:h4 (stylefy/use-style style-form-fields/checkbox-group-label) label])
+     (when help
+       [common/help help])
+     (doall
+       (map-indexed
+         (fn [i option]
+           (let [checked? (boolean (selected option))]
+             ^{:key i}
+             [:div {:style {:display "flex" :flex-wrap "wrap" :justify-content "flex-start" :align-items "center" :padding-top "10px"}}
```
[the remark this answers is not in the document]
rckopteri: This overrides material ui styles. If I use class from stylefy it won't work. So this was the best solution I could think of.
But I'll change those px to rem

ote/src/cljs/ote/app/controller/transport_operator.cljs
rckopteri: thanx

ote/src/cljs/ote/views/transport_operator_ytj.cljs
rckopteri: Yes. It is possible. And I didn't even consider this option when I was working with this. And passing a function might take less memory (not significant). But I think that right now I leave this as it is even it might be better solution... I'm not sure if it is. It might be and probably is but still :)

PR #800, opened 2019-01-08
162754989 give permissions to all users

rckopteri:
# Added
* Give permissions to all users within the same business-id family

### DIFF
--- a/ote/src/clj/ote/services/transport.clj
+++ b/ote/src/clj/ote/services/transport.clj
@@ -177,7 +177,7 @@
         member (create-member! db user-id group)
         ;; Ensure that other users get permissions to new operator as well
         ;; Get other possible users that need permissions
-        other-users (fetch-users-with-same-business-id db {:business-id (::t-operator/business-id op)
+        other-users (fetch-users-within-same-business-id-family db {:business-id (::t-operator/business-id op)
                                                            :user-id user-id})]
     (when other-users
       (doall
--- a/ote/src/clj/ote/services/transport.clj
+++ b/ote/src/clj/ote/services/transport.clj
@@ -150,11 +150,21 @@
      :transport-service-vector transport-services-vector
      :user cleaned-user}))
 
-(defn- create-group!
+(defn- create-member! [db user-id group]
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/services/transport.clj
```diff
@@ -164,13 +174,16 @@
...
-                                ::user/group_id   (:ote.db.transport-operator/group-id group)
-                                ::user/table_name "user"
-                                ::user/capacity   "admin"
-                                ::user/state      "active"})]
+        member (create-member! db user-id group)
+        ;; Ensure that other users get permissions to new operator as well
+        ;; Get other possible users that need permissions
+        other-users (fetch-users-within-same-business-id-family db {:business-id (::t-operator/business-id op)
+                                                           :user-id user-id})]
+    (when other-users
```
philosotractor: Would following (or similar if this doesn't work) eliminate the need for explicit `when` and `doall`?
```(for [usr other-users]
    (create-member! db (:user-id usr) group))```

File: ote/src/clj/ote/services/transport.clj [hunk 2]
```diff
@@ -164,13 +174,16 @@
...
                                ::t-operator/approval_status "approved"
                                ::t-operator/is_organization true})
-        member (specql/insert! db ::user/member
-                               {::user/id         (str (UUID/randomUUID))
-                                ::user/table_id   (get-in user [:user :id])
-                                ::user/group_id   (:ote.db.transport-operator/group-id group)
-                                ::user/table_name "user"
-                                ::user/capacity   "admin"
-                                ::user/state      "active"})]
+        member (create-member! db user-id group)
```
philosotractor: Would removing the sql `WHERE` condition "```AND u.id != :user-id```" allow removing this separate adding of user to own organisation, would the loop below then add also himself?

rckopteri: I had a reason to do it this way, but now when I look at this it seems stupid and I don't remember why I needed to do this using loop for others and direct call for the creator. Jeah. I'll change this. Thanx

PR #805, opened 2019-01-09
Monitor view (aka valvontankymä) CSV download & view explanation texts

erno:
See #162907408


ote/src/clj/ote/services/reports.sql
```diff
@@ -171,10 +171,9 @@ select
 
 -- name: operator-type-distribution
 -- returns a distrubution of transport-service sub-types among all transport services
-select
-  "sub-type",
-  count("sub-type") as count
-  from (select distinct top.id as toid,ts."sub-type" from "transport-service" ts, "transport-operator" top where top.id = ts."transport-operator-id" and ts."sub-type" is not null) as unusedname
+select "sub-type", count("sub-type") as count
```
philosotractor: Code style in this and previous `select`s are different, file style is uppercase for SQL keywords.

PR #819, opened 2019-01-15
15708517 service search

jkorpelin:
# Fixed
* Translations.
   
# Changed
* Changed service search to include associated services.
   
# Datamodel
* Changed service search view.


database/src/main/resources/db/migration/V1_128__service_search_view_change.sql
```diff
@@ -0,0 +1,26 @@
+-- Drop old
+DROP VIEW transport_service_search_result;
+
+-- Create new with t.companies in service-companies
```
rckopteri: create new with "associated-service-operators"

database/src/main/resources/db/migration/V1_129__associated_service_operators_change.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "associated-service-operators"
+  ADD COLUMN timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  ADD COLUMN "user-id" TEXT REFERENCES public.user (id);
```
rckopteri: public isn't needed. Would be better to leave it out so the style remains the same

PR #822, opened 2019-01-17
163220298 transition in delete operator

rckopteri:
# Changed
* Transition from operator page to own services when operator is deleted. Previously we moved user away from operator page always when operator was deleted. Now we change page only if non-ytj company is deleted.


philosotractor:
Looks good, if it works then go ahead.

However.

In this PR the three operator vectors 
(`transport-operators-to-save`, `transport-operators-with-services`,  `ytj-company-names`) 
are updated directly to remove the deleted nap operator. 
An alternative approach could be to invoke 
`refresh-transport-operator` which results into `OperatorRefreshResponse`,  which in turn already updates the vectors.

PR #826, opened 2019-01-17
15708517 tests

jkorpelin:
# Added
* new cypress tests
* Way to pass a data attribute for chip-input


ote/src/cljs/ote/ui/form_fields.cljs
```diff
@@ -380,6 +380,7 @@
         :hintText (or hint-text (placeholder field data))
         :disabled disabled?
         :value chips
+        :data-cypress data-attribute-cypress
```
rckopteri: What if data-attribute-cypress is not given? It might be better to add this attribute only if it is given in form)

cypress/integration/own-services_spec.js
```diff
@@ -186,5 +186,33 @@ describe('Add new service provider', function () {
 
         cy.contains('Palveluntuottaja poistettiin onnistuneesti.');
     });
+});
+
+describe('Should add new associated service to Normal users "Terminaali Oy" and delete them', function () {
+    before(function() {
+        cy.login('normaluser', 'password')
```
philosotractor: Does this need to define default user explicitly? 
Here test could call the custom command `cy.login()` which resolves the username pwd itself.

ote/src/cljs/ote/views/own_services.cljs
```diff
@@ -247,7 +247,8 @@
    [:ul.unstyled
     (doall
       (for [as oa-services]
-        [:li {:key (:service-id as)}
+        [:li {:key (:service-id as)
+              :data-cypress-op-id (:service-id as)}
```
philosotractor: Wouldn't `:id` suite better?
Also, DOM doesn't have to refer to particular tool technology solution is implemented with so the `cypress` part is a bit striking naming wise. `:id`could be used by any functionality, not only e2e tests.

cypress/integration/own-services_spec.js
```diff
@@ -186,5 +186,33 @@ describe('Add new service provider', function () {
...
+        cy.visit('/#/own-services');
+    });
+
+    it('Should be in own-services page', function () {
+        cy.contains('Omat palvelutiedot');
+    });
+
+    it('Should add a new associated service', function () {
+        cy.get('input[data-cypress=chip-input]').type('t');
+        cy.contains('Taksi - Ajopalvelu Testinen').click();
```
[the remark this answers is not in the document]
philosotractor: For maintainability woudl be better to declare test data in an object, which is then referred in tests. Or similar strucure.

ote/src/cljs/ote/views/own_services.cljs
jkorpelin: https://docs.cypress.io/guides/references/best-practices.html according to cypress best practices using data attribute is the best because it's insulated from all changes.

philosotractor: Great that this followed the tool convention, 
although I don't agree with this
"Your application may use dynamic classes or ID’s that change".  
When we're giging the id, we control if it changes or not. 
Semantically an "id" or "name" attribute sound more proper way on how to selecxt things than a "data". We just need to agree what convention we use :)

cypress/integration/own-services_spec.js
jkorpelin: Added cy.normalLogin() function to login with normaluser account.

ote/src/cljs/ote/views/own_services.cljs
jkorpelin: I still think it's better to have data-attributes since with them, you can ben **100%** sure that they only affect tests. With id's you could in theory have css that targets that tag.

ote/config.edn
```diff
@@ -9,7 +10,7 @@
         :auth-tkt {:shared-secret "localdev"
                    :max-age-in-seconds 36000
                    :digest-algorithm "MD5"}
-        :session {:key "cookie0123456789"}
```
philosotractor: Accidental change?

ote/config.edn
```diff
@@ -1,5 +1,6 @@
 {:environment {:base-url "http://localhost:3000/"}
  :dev-mode? true
+ :production-env? false
```
philosotractor: Accidental change?

ote/src/cljs/ote/ui/buttons.cljs
```diff
@@ -45,3 +46,10 @@
    (common/linkify url
                    [ui/flat-button {:label label :primary true}]
                    {:target "_blank"})])
+
+(defn icon-button
+  [opts label]
+  [:button.unstyled-button (merge
+             opts
```
philosotractor: I wonder, should `opts` be last in `merge` to allow overwriting the default style from stylefy call?

cypress/integration/own-services_spec.js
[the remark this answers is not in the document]
jkorpelin: I think it's premature to add it now since the current values would just be behind another variable and it would make it more difficult to change.

PR #827, opened 2019-01-17
163208379 operator wizard selections

philosotractor:
# Changed
* transport-operator wizard section (for selecting ytj name for those nap
operators which do not have a ytj name match), supports now
an empty default value and selecting & reverting selection per orphan.
And updating operators-to-save checkbox list accordingly.

### DIFF
--- a/ote/src/cljs/ote/ui/select_field.cljs
+++ b/ote/src/cljs/ote/ui/select_field.cljs
@@ -37,7 +37,7 @@
          (map-indexed
            (fn [i option]
              (if (= :divider option)
-               ^{:key (str "select-field-divider")}
+               ^{:key (str "select-field-divider-" i)}
                [ui/divider]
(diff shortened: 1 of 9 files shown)


ote/src/cljs/ote/ui/select_field.cljs
```diff
@@ -33,9 +37,15 @@
          (map-indexed
            (fn [i option]
              (if (= :divider option)
-               ^{:key i}
+               ^{:key (str "select-field-divider")}
```
rckopteri: If map-indexed has more than one element, key is not unique?

philosotractor: True, `i` was supposed to be appended. Will fix, thanks.

philosotractor: Although above applies if there are several `:divider`s

PR #828, opened 2019-01-18
163015263 brokering service dialog

rckopteri:
# Added
* Confirm dialog for brokering services

### DIFF
--- a/ote/src/cljs/ote/app/controller/transport_service.cljs
+++ b/ote/src/cljs/ote/app/controller/transport_service.cljs
@@ -58,10 +58,9 @@
   {}
   (let [type (get-in app [:transport-service ::t-service/type])
         type-key (t-service/service-key-by-type type)]
-  (cond-> app
-          (when (not select-type)) (assoc :show-brokering-service-dialog? false)
-          true (assoc-in [:transport-service :show-brokering-service-dialog?] false )
-          true (assoc-in [:transport-service type-key ::t-service/brokerage?] select-type))))
+  (-> app
+      (assoc-in [:transport-service :show-brokering-service-dialog?] false )
+      (assoc-in [:transport-service type-key ::t-service/brokerage?] select-type))))
 
 ;;; Navigation hook events for new service creation and editing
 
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/app/controller/transport_service.cljs
```diff
@@ -50,6 +50,19 @@
...
+(define-event ShowBrokeringServiceDialog []
+  {}
+  (assoc-in app [:transport-service :show-brokering-service-dialog?] true))
+
+(define-event SelectBrokeringService [select-type]
+  {}
+  (let [type (get-in app [:transport-service ::t-service/type])
+        type-key (t-service/service-key-by-type type)]
+  (cond-> app
+          (when (not select-type)) (assoc :show-brokering-service-dialog? false)
```
philosotractor: In `cond` the first arg is a boolean true/false whether the next element should be evaluated or not. While `when` probably works because it returns either a alue or nil, it might be more logical to simply use a boolean evaluation  i.e. ```(not select-type)´´´

rckopteri: Really good point. thanx!

ote/src/cljs/ote/views/transport_service_common.cljs
```diff
@@ -400,28 +402,50 @@
...
+                       :primary true
+                       :on-click #(e! (ts/->SelectBrokeringService true))}])]}
+        [:p (tr [:dialog :brokering-service :body])]])
+
+     ;show-footer? - Take owner check away for now
+     (when true
+       [:div.row
+        (when (not (form/can-save? data))
+          [ui/card {:style {:margin-bottom "1rem"}}
+           [ui/card-text {:style {:color "#be0000" :padding-bottom "0.6rem"}} (tr [:form-help :publish-missing-required])]])
```
philosotractor: Colour codes should be `def`'d in `colors.clj` and flow through a semantical `def`.
For an example please take a look on `(def warning red-basic)`.

Similar for the `:padding-bottom`.

All in all a recommended pattern could be here to have a view or card-specific `def` style in a new transport-service.cljs or append to base.cljs, then just apply that def´d style here to the card element.
That style then would be composed like comments above propose.
Result would be that updating styles via def is straightforward and maintainable.

rckopteri: Agreed. However I'm working on a branch with several ui /warning fixes and I'll take care of this in that branch.
(this is old code and I didn't change it in this pr)

PR #833, opened 2019-01-22
163361831 compare days 500 error

rckopteri:
# Fixed
* Changed gtfs_route_difference stored procedure to work with route-hash-id only.
* This prevents finding multiple route differences and fixes the bug.
* Also added few random numbers to table component to prevent js console warnings. However our table component doesn't take into account :keys that are passed to header row it is always 'h1' and it cannot be fixed using this version of material ui.

### DIFF
--- a/database/src/main/resources/db/migration/R__GTFS_query.sql
+++ b/database/src/main/resources/db/migration/R__GTFS_query.sql
@@ -292,8 +292,7 @@ BEGIN
 END
 $$ LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION gtfs_route_differences(
-     "route-short-name" TEXT, "route-long-name" TEXT, "trip-headsign" TEXT, "route-hash-id" TEXT,
+CREATE OR REPLACE FUNCTION gtfs_route_differences("route-hash-id" TEXT,
      d1_trips "gtfs-package-trip-info"[], d2_trips "gtfs-package-trip-info"[])
 RETURNS "gtfs-route-change-info"
 AS $$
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/ui/table.cljs
```diff
@@ -19,6 +19,7 @@
                      label-style
                      row-style show-row-hover?
                      on-select row-selected? no-rows-message class] :as opts} headers rows]
+  (let [random-table-id (str "tid-"(rand-int 999))]
```
jkorpelin: Maybe (java.util.UUID/randomUUID) instead of rand-int? Lower chance of collision.

philosotractor: Or id via `range` to have no chance of collision?

philosotractor: For the sake of discussion, another alternative is to pass the table :key value from view who is creating the table, to here. View could assign unique keys using a string and an iterated integer.

database/src/main/resources/db/migration/R__GTFS_query.sql
```diff
@@ -392,9 +391,9 @@ BEGIN
 
   --RAISE NOTICE 'yli jäi d1: % ja d2: % vuoroa', array_length(d1_trip_ids, 1), array_length(d2_trip_ids, 2);
 
-  route_changes."route-short-name" := "route-short-name";
-  route_changes."route-long-name" := "route-long-name";
-  route_changes."trip-headsign" := "trip-headsign";
+  route_changes."route-short-name" := null; --"route-short-name";
```
philosotractor: Are the old values left in comments on purpose?

rckopteri: route_changes uses composition type which requires something to be added on those columns.
When we migrate to composite type to own table, this won't be needed anymore. So yes and no :) we don't need those comments but this is going to change soon anyway.

ote/src/cljs/ote/ui/table.cljs
rckopteri: I leave it as it is. We have max 3 tables on same page so there is very low change of collision. We have a ticket for creating new table component and I don't want to spend any more time fixing this old one.

PR #839, opened 2019-01-23
161836698 button styles updated globally

philosotractor:
# Added
* New styles for primary, negative, outline buttons and related elements like text colour
* New negative (=red) button
   
# Fixed
* Some problem fixed.
   
# Changed
* Changed following buttons to use html button element instead of material ui components: save, delete, cancel 
* buttons updated for buttons 
 - embedded in tables a few old ones like file upload remain). Tables e.g. in service and sea route views
 - navigation-prompt
 - service/operator/(sea)route deletion modal
 - user settings
 - user email settigs
 - user login
 - user registration

### DIFF
--- a/ote/src/cljs/ote/style/base.cljs
+++ b/ote/src/cljs/ote/style/base.cljs
@@ -32,14 +32,6 @@
 (def icon-large {:width 32 :height 32
                  :vertical-align "middle"})
 
-(def base-button {:padding-left "1.1em"
-                  :padding-right "1.1em"
-                  :text-transform "uppercase"
-                  :color "#FFFFFF"
-                  :background-color "#1565C0"
-                  :font-size "12px"
-                  :font-weight "bold"})
-
 (def blue-link-with-icon {:color colors/primary
                           :text-decoration "none"
                           :margin-bottom "1px"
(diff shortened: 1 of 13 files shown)


ote/src/cljs/ote/style/base.cljs
```diff
@@ -51,9 +43,21 @@
...
 
+(def base-button (merge
+                   {:padding-left "1.1em"
+                         :padding-right "1.1em"
+                         :text-transform "uppercase"
+                         :color "#FFFFFF"
+                         :background-color "#1565C0"
+                         :font-size "12px"
+                         :font-weight "bold"}
+                   blue-link-with-icon))
```
jkorpelin: This feels unnecessary. "base button" style will now allways have border button on hover?

ote/src/cljs/ote/style/buttons.cljs
```diff
@@ -16,23 +21,38 @@
...
+   :background-color colors/negative-button-hover
+   :transform "scale(0.98)"})
+
 (def button-common
   {:padding "20px"
-   :height "60px"                                           ;;Hard coded so the height doesn't change with transitions
+   :height "60px"                                           ;; Hard coded so the height doesn't change with transitions
+   :min-width "4rem"
    :display "inline-block"
+   :justify-content "center"
```
jkorpelin: This only works with display "flex" elements. Should use text-align "center"

philosotractor: True, will change. Thanks!

ote/src/cljs/ote/style/base.cljs
philosotractor: Some WIP code which should be removed, will do.

philosotractor:
Resolved some master conflicts and force pushed everything 
commit 342b0c3 fixes @jkorpelin 's remarks.

ote/src/cljs/ote/style/base.cljs
```diff
@@ -51,9 +43,18 @@
                           ::stylefy/mode {:hover {:border-bottom (str "1px solid " colors/primary)
                                                   :margin-bottom "-1px"}}})
 
+(def base-button
```
rckopteri: I would move these to buttons styles file

philosotractor: Yes, but `base-button` is an old style which seems to be used only in one place anymore. These commits did not modify `base-button`.

It seems to be about the delete-button (bin icon) in service tables. I order to finish things, this task will update only action buttons and future tasks may improve layouts and styles more :)

PR #852, opened 2019-01-30
Transport service catalog change

jkorpelin:
# Changed
* Transport service catalog -> removed all but the current searched provicder from the "other companies providing the service" list.
   
# Datamodel
* Changed transport_service_search_result view to contain all the participating service provicers properly.

### DIFF
--- a/database/src/main/resources/db/migration/V1_130__service_search_view_fix.sql
+++ b/database/src/main/resources/db/migration/V1_130__service_search_view_fix.sql
@@ -1,0 +1,29 @@
+-- Drop old view
+DROP VIEW transport_service_search_result;
+
+-- Create new with "associated-service-operators"
+CREATE VIEW transport_service_search_result AS
+  SELECT t.*, op.name as "operator-name", op."business-id" as "business-id",
+         array_cat(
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/services/service_search.clj
```diff
@@ -153,7 +153,16 @@
                         (update-in result
                                    [::t-service/external-interface-links]
                                    #(mapv hide-import-errors %)))
-                      results)]
+                      results)
+        results (mapv
+                  #(update % ::t-service/service-companies
+                           (fn [c]
+                             (do
```
rckopteri: do is not needed anymore.

database/src/main/resources/db/migration/V1_130__service_search_view_fix.sql
```diff
@@ -0,0 +1,29 @@
+-- Drop old view
+DROP VIEW transport_service_search_result;
+
+-- Create new with "associated-service-operators"
+CREATE VIEW transport_service_search_result AS
```
rckopteri: Could you check intentions if there is a way to improve them

PR #855, opened 2019-01-31
163619228 route differences

rckopteri:
# Fixed
* 500 error when fetching route differences (which are trip & stop differences actually)
* I started using same implementation that is used in detection. So the stored procedure is now unused and I removed it completely.
* This Clojure version is easier to debug, understand and change. And it is used from two places now. It also as fast as the stored procedure so the change is win win.


ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -175,7 +175,7 @@
 
 
 (defn combine-trips [date1-trips date2-trips]
-  (when (and date1-trips date2-trips)
+  (when (and date1-trips date2-trips (> (count date1-trips) 0) (> (count date2-trips) 0))
```
philosotractor: `>` could be replaced using `pos-int?`

PR #856, opened 2019-02-01
Changed how information for services and providers is shown.

jkorpelin:
Also removed operators route from service and view from frontend.

   
# Changed
* How information for services and providers is shown.
* removed backend route for services


ote/src/clj/ote/services/service_search.clj
```diff
@@ -161,10 +161,14 @@
                                        (when (and operators (::t-service/business-id company))
                                          (.contains operators (::t-service/business-id company))))
                                      c)))
-                  results)]
+                  results)
+        results-without-personal-info (mapv
+                                        (fn [result]
+                                          (dissoc result ::t-service/contact-email ::t-service/contact-address ::t-service/contact-phone))
```
rckopteri: This dissoc and the dissoc in geojson.clj are the same. They could be one function instead of copy&paste

PR #857, opened 2019-02-01
Add service-id to gtfs-date-hash table

rckopteri:
# Datamodel
* Add service-id to gtfs-date-hash table to speed up problem solving


database/src/main/resources/db/migration/V1_131__add_service_id_to_date_hash.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "gtfs-date-hash"
+  ADD COLUMN "transport-service-id" INTEGER;
```
philosotractor: LGTM
Just double checking do you think if foreign key should be used or not?
`package-id` is already a FK, which is pretty unambiguous so I suppose that's enough already.

PR #860, opened 2019-02-04
163649260 table component

jkorpelin:
# Added
* Tables show different outline for selected rows
   
# Fixed
* Some icon colors not being visible on firefox.


ote/src/cljs/ote/theme/colors.cljs
```diff
@@ -2,6 +2,7 @@
 
 (def blue "#0048c2")
 (def blue-light "#0052dd")
+(def blue-dark "#0029B8")
```
philosotractor: While this probably looks ok, I now see current colour codes for blue don't match to UI spec colour codes. 
Eg. #0048c2/rgb(0,102,204) is `blue-dark`.
Would be good to alighn these in a future PR. I think the other like gray/red/orange are already aligned.

ote/src/cljs/ote/ui/table.cljs
```diff
@@ -17,86 +19,88 @@
...
-      (doall
-        (map-indexed
-          (fn [i row]
-            ^{:key (if key-fn (key-fn row) (str i "-" random-table-id "-body-row"))}
-            [ui/table-row {:selectable     (boolean on-select)
-                           :selected       (if row-selected? (row-selected? row) false)
-                           :display-border false
+    [ui/table (merge
+                {:wrapperStyle {:overflow "visible"}
+                 :body-style {:padding-right "5px"}
```
philosotractor: Was a fraction of `rem`considered here? No big reason to ask, just that relative units should be used unless there's a problem with them in a use-case.

ote/src/cljs/ote/ui/table.cljs
```diff
@@ -17,86 +19,88 @@
...
+     [ui/table-body {:display-row-checkbox false}
+      (if (empty? rows)
+        [:tr [:td {:colSpan (count headers)}
+              [common/help no-rows-message]]]
+        (doall
+          (map-indexed
+            (fn [i row]
+              ^{:key (if key-fn (key-fn row) (str i "-" random-table-id "-body-row"))}
+              [ui/table-row (merge
+                              (stylefy/use-style (merge {:background-color (if (= 0 (rem i 2)) "#FFFFFF" (color :grey200))} ; Add stripes
```
philosotractor: Instead of 
`(if (= 0 (rem i 2)) "#FFFFFF" (color :grey200))`
would following be more readable?
`(if (even? i) "#FFFFFF" (color :grey200))`

Ps. Also in case of zero checking `zero?` should be preferred :)

philosotractor: Should colour code be`:gray200` instead of `:grey200` ?

Also, should table use a logical colour symbol like `table-row-color-alt` and `table-row-color`, which are in turn defined to `gray200` or whatever?

ote/src/cljs/ote/theme/colors.cljs
jkorpelin: This color was given by Jussi, so I don't think the spec is up to date.

philosotractor: Ok these shoudl be double checked with him. Previously it has been discussed that he too takes the codes from the spec, and so should we.

ote/src/cljs/ote/ui/table.cljs
jkorpelin: Good point. This was old code just moved around. Fixed now.

PR #862, opened 2019-02-05
163446864 change date discrepancy fix

erno:
A week with a following no-traffic week as the next+1 week used to always read as not-changed. Now the week= comparison fn will special case a fully empty week.


erno:
Still WIP as testing ran into a NPE and calendar view doesn't really look happy either on dev setup.
Probably the all-nil week vector is getting into some code paths that are not handling it correctly.

PR #866, opened 2019-02-07
163478330 service card update

jkorpelin:
# Added
* New service card for service search
* Breakpoints for buttons to go smaller


ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -90,26 +92,22 @@
...
+   [:div
+    [:span (str (tr [:service-search :service-basic-info]) " | GeoJSON")]]
+   (when-not (empty? external-interface-links)
+     (doall
+       (map-indexed
+         (fn [i {::t-service/keys [external-interface format data-content] :as row}]
+           (let [data-content (if (nil? data-content)
+                                (::t-service/url external-interface)
+                                (parse-content-value data-content))]
+             (do
```
philosotractor: This `do` should be possible to be removed

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -225,7 +224,8 @@
        (doall
          (for [result results]
            ^{:key (::t-service/id result)}
-           [result-card e! (:admin? user) result service-search (:width app)]))
+           [:div
```
philosotractor: Did you check if this container div is needed or not? Just a thought.

jkorpelin: True accidentally left it there when i had both new and old card styles visible. Fixed

PR #869, opened 2019-02-08
Fixes empty routes appearing in transit visualization.

jkorpelin:
# Fixed
* Empty row in route changes table.
   
# Changed
* `gtfs_service_routes_with_daterage` to check if the hash id is valid.


database/src/main/resources/db/migration/R__GTFS_query.sql
```diff
@@ -532,6 +532,8 @@ SELECT COALESCE(rh."route-short-name",'') AS "route-short-name",
   -- Join unnested per route hashes
   JOIN LATERAL unnest(dh."route-hashes") rh ON TRUE
  WHERE rh.hash IS NOT NULL
+  AND rh."route-hash-id" IS NOT NULL
```
rckopteri: Just a note. River formatting would be nice. (you can refactor this in another pr)

PR #875, opened 2019-02-13
163788230 transit-visualization filter checkbox fix if data does not change

philosotractor:
# Fixed
* views,controller: transit-visualization filter checkbox fix if data does not change
Otherwise checkbox remains disabled if route-changes model does not change
and component-did-mount event never clears the flag used to disable
control.

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -473,6 +473,8 @@
   ;; Disabling of UI components must happen before table model change because otherwise table rendering
   ;; delays those as well.
   (.setTimeout js/window #(e! (->ToggleShowNoChangeRoutesDelayed)) 0)
-  (-> app
-      (update :show-no-change-routes-checkbox? not)
-      (assoc :route-changes-loading? true)))
+  (cond-> app
+      true (update :show-no-change-routes-checkbox? not)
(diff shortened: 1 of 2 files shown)


philosotractor:
While this works, perhaps this is too complex solution compared to the problem it solves.
I'll consider other alternatives like removing the disabling alltogether.

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -473,6 +473,8 @@
   ;; Disabling of UI components must happen before table model change because otherwise table rendering
   ;; delays those as well.
   (.setTimeout js/window #(e! (->ToggleShowNoChangeRoutesDelayed)) 0)
-  (-> app
-      (update :show-no-change-routes-checkbox? not)
-      (assoc :route-changes-loading? true)))
+  (cond-> app
+      true (update :show-no-change-routes-checkbox? not)
+      ;; :route-changes-loading? set only when table model changes, otherwise there's no :component-did-mount event to clear the flag
+      (not= (:gtfs/route-changes (:changes-no-change app))
```
mattiuusitalo: Opportunity to use get-in [https://clojuredocs.org/clojure.core/get-in] ?

philosotractor:
Closed because decided for an alternative approach. Continues in
https://github.com/finnishtransportagency/mmtis-national-access-point/pull/891

PR #878, opened 2019-02-14
163941143: main header :user-menu-video-tutorials-link commented out

philosotractor:
# Changed
* main header :user-menu-video-tutorials-link commented out. For now videos are not available. Shall be restored when they are.

### DIFF
--- a/ote/src/cljs/ote/ui/main_header.cljs
+++ b/ote/src/cljs/ote/ui/main_header.cljs
@@ -216,11 +216,13 @@
            (merge (stylefy/use-style
                     style-topnav/topnav-dropdown-link)
                   {:target "_blank"})]]
-         [:li
-          [linkify (tr [:common-texts :user-menu-video-tutorials-link]) (tr [:common-texts :user-menu-video-tutorials])
-           (merge (stylefy/use-style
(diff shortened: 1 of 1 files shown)


ote/src/cljs/ote/ui/main_header.cljs
```diff
@@ -216,11 +216,13 @@
...
                     style-topnav/topnav-dropdown-link)
                   {:target "_blank"})]]
-         [:li
-          [linkify (tr [:common-texts :user-menu-video-tutorials-link]) (tr [:common-texts :user-menu-video-tutorials])
-           (merge (stylefy/use-style
-                    style-topnav/topnav-dropdown-link)
-                  {:target "_blank"})]]
+
+         ;; TODO: commented out because for now there are no valid videos. Link shall be restored and updated when videos are available.
+         ;[:li
```
rckopteri: #_ would have do the trick :)

philosotractor: True. In some cases I prefer row comments, because then one sees already in the "find in files" one row preview result if row is commented. Also, toggling row-comments in batch via IDE hotkey is one click :)

PR #886, opened 2019-02-19
Silence Hikari stats debug output in REPL

erno:


ote/src/clj/ote/main.clj
```diff
@@ -120,6 +120,11 @@
 (defn configure-logging [dev-mode? {:keys [level] :as log-config}]
   (log/merge-config!
    {:level (or level :debug)
+    :middleware [(fn drop-hikari-stats-middlware [data]
```
rckopteri: nice!

PR #887, opened 2019-02-20
164075691 routes without changes missing

philosotractor:
# Changed
* services: remove unused changed-routes-to-service-by-date sql
* views: transit-visualization fix route-change checkbox disabling
* services: transit-visualization fix for route-changes and keyword type, not-changes routes were not shown
* util: file for clojure-language related functions with remove-coll-key-ns

### DIFF
--- a/ote/src/clj/ote/util/clojure.clj
+++ b/ote/src/clj/ote/util/clojure.clj
@@ -1,0 +1,4 @@
+(ns ote.util.clojure)
+
+(defn remove-coll-key-ns [c]
+  (into {} (map (fn [[k v]] [(keyword (name k)) v]) c)))
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -24,9 +24,9 @@
    "#ec8fb5" "#23dbe1" "#a4515b" "#169294" "#fd5925" "#3d4e92" "#f4d403"
    "#66a1e5" "#d07d09" "#9382e9" "#b9cf84" "#544437" "#f2cdb9"])
 
-(defn route-filtering-available? [{:keys [changes-filtered changes-no-change route-changes-loading?] :as transit-visualization}]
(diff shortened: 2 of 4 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -87,7 +87,7 @@
   "Sort route changes according to change date and route-long-name: Earliest first and missing date last."
   [show-no-change changes]
   (let [;; Removed in past routes won't be displayed at the moment. They are ended routes and we do not need to list them.
-        removed-in-past (sort-by (juxt :route-long-name :route-short-name) (filterv #(and (= :removed (:change-type %)) (nil? (:change-date %))) changes))
+        ;removed-in-past (sort-by (juxt :route-long-name :route-short-name) (filterv #(and (= :removed (:change-type %)) (nil? (:change-date %))) changes))
```
mattiuusitalo: Is it necessary to save this commented out code?

ote/src/clj/ote/util/clojure.clj
```diff
@@ -0,0 +1,4 @@
+(ns ote.util.clojure)
+
+(defn remove-coll-key-ns [c]
+  (into {} (map (fn [[k v]] [(keyword (name k)) v]) c)))
```
rckopteri: No need to change now, but next time this would be easier to understand and change
(into {}
  (map 
    (fn [[key value]] 
     [(keyword (name key)) value]) 
   c)))

ote/src/cljs/ote/app/controller/transit_visualization.cljs
philosotractor: Very good question, previously it was _unused_ and here it's just commented. Previous comment explains why, but I didn't dare to remove it as part of this yet.

ote/src/clj/ote/util/clojure.clj
philosotractor: True :)

PR #900, opened 2019-02-22
163649701 skip contract traffic

rckopteri:
# Added
* New tab to admin panel where admin users can mark if scheduled service is commercial or contract traffic.

### DIFF
--- a/ote/src/clj/ote/services/admin.clj
+++ b/ote/src/clj/ote/services/admin.clj
@@ -331,6 +331,14 @@
                 {::t-operator/business-id business-id})]
     (map #(assoc {} :transport-operator %) result)))
 
+(defn- get-commercial-scheduled-services [db]
+  (fetch-commercial-services db))
+
+(defn- toggle-commercial-service [db {id :id commercial? :commercial? :as form-data}]
+  (specql/update! db ::t-service/transport-service
+                  {::t-service/commercial-traffic? commercial?}
+                  {::t-service/id id}))
+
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/app/controller/admin.cljs
```diff
@@ -560,3 +561,42 @@
...
+              :on-failure (tuck/send-async! ->ServerError)})
+  app)
+
+(defmethod routes/on-navigate-event :admin-commercial-services [_]
+  (->LoadCommercialServices))
+
+(define-event ToggleCommercialTrafficResponse [response]
+  {}
+  ;; Do nothing on success
+  (.log js/console "Response " (pr-str response))
```
jkorpelin: This is apparently here on purpose?

rckopteri: yes

ote/src/clj/ote/services/admin.clj
```diff
@@ -331,6 +331,14 @@
                 {::t-operator/business-id business-id})]
     (map #(assoc {} :transport-operator %) result)))
 
+(defn- get-commercial-scheduled-services [db]
+  (fetch-commercial-services db))
+
+(defn- toggle-commercial-service [db {id :id commercial? :commercial? :as form-data}]
```
jkorpelin: Is the `:as form-data` needed when the form-data isn't used?

PR #901, opened 2019-02-22
163934214 Clear selected route when opening transit visualization

mattiuusitalo:
# Fixed
Previously:
1. given user had selected a service in "tunnistetut muutokset"
2. and selected a route
3. when user navigated back to "tunnistetut muutokset"
4. and selected another service
then the route selected in step 2 was still visible in transit visualization view

Now the selected route is cleared from app state when user navigates to transit visualization view.

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -104,7 +104,8 @@
                       (dissoc :changes-route-no-change)
                       (assoc :service-changes-for-date-loading? true)
                       (assoc :all-route-changes-display? false)
-                      (assoc-in [:open-sections :gtfs-package-info] false)))]
+                      (assoc-in [:open-sections :gtfs-package-info] false)
+                      (dissoc :selected-route)))]
     (update app :transit-visualization init-tv)))
 
 (define-event RoutesForDatesResponse [routes dates]
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -95,19 +95,6 @@
                              sorted-changes)]
     all-sorted-changes))
 
-(defn- init-view-state [app]
-  (let [init-tv (fn [transit-visualization]
-                  (-> transit-visualization
-                      (assoc :all-route-changes-checkbox nil)
-                      (assoc :all-route-changes-display? nil)
-                      (dissoc :changes-route-filtered)
-                      (dissoc :changes-route-no-change)
-                      (assoc :service-changes-for-date-loading? true)
-                      (assoc :all-route-changes-display? false)
-                      (assoc-in [:open-sections :gtfs-package-info] false)
-                      (dissoc :selected-route)))]
-    (update app :transit-visualization init-tv)))
-
 (define-event RoutesForDatesResponse [routes dates]
   {:path [:transit-visualization :compare]}
   (if (= dates (select-keys app [:date1 :date2]))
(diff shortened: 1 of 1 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -104,7 +104,8 @@
                       (dissoc :changes-route-no-change)
                       (assoc :service-changes-for-date-loading? true)
                       (assoc :all-route-changes-display? false)
-                      (assoc-in [:open-sections :gtfs-package-info] false)))]
+                      (assoc-in [:open-sections :gtfs-package-info] false)
+                      (dissoc :selected-route)))]
```
rckopteri: I think that this will fix the issue, but I was wondering if we could clean up that :transit-visualization key completely?
Something like this

(fn [transit-visualization]
                  (-> transit-visualization	                  (-> transit-visualization
                      (assoc :all-route-changes-checkbox nil)	                      (assoc :all-route-changes-checkbox nil)
                      (assoc :all-route-changes-display? nil)	                      (assoc :all-route-changes-display? nil)                                        
                      (assoc :service-changes-for-date-loading? true)	                      (assoc :service-changes-for-date-loading? true)
                      (assoc :all-route-changes-display? false)	                      (assoc :all-route-changes-display? false)
                      (assoc-in [:open-sections :gtfs-package-info] false)))]	                      (assoc-in [:open-sections :gtfs-package-info] false)))

(assoc app :transit-visualization init-tv)

philosotractor: Can't make this out, was there a copy-paste problem here? @rckopteri 

I agree that cleaning all keys should be done whenever entering a view. Whether to `dissoc` or `nil` them is a matter of how logic is implemented, but generally speaking dissocing sounds better if there's no reason to do otherwise.

mattiuusitalo: Let's try if this helps. @jkorpelin noticed that on a slower machine / browser there are still some glitches when navigating to the view a second time.

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -187,6 +175,13 @@
                 :gtfs-package-info (:gtfs-package-info response)
                 :route-hash-id-type (:route-hash-id-type response)))
 
+(defn- init-view-state [app]
+  (let [initial-view-state {:all-route-changes-checkbox nil
+                            :all-route-changes-display? false
+                            :open-sections {:gtfs-package-info false}
+                            :service-changes-for-date-loading? true}]
+    (assoc app :transit-visualization initial-view-state)))
+
```
rckopteri: This is good. But you still could do one more thing.

In transit-changes.clj controller at row 30 forwards you could add :transit-visualization clean up.
Like so:
Replace {:path [:transit-changes]} -> {}

And from row 34
 (-> app
      (assoc-in [:transit-changes :loading?] true)
      (assoc-in [:transit-changes :show-all] true)
      (dissoc :transit-visualization))

mattiuusitalo: That would indeed reset transit-visualization when navigating to this view via transit-changes. Problem is, you potentially can reah transit-visualization from other views and this fix does not cover that. As a general note transit-visualization related functionality should not leak into other components.

philosotractor: +1
Generic design should be that a view initializes itself upon navigation and doesn't rely on other views - if it does, then it should enforce to prevent other navigation which might break the workflow. Which is not the case here :)

PR #906, opened 2019-02-25
163945472 calculation status

rckopteri:
# Added
* admin: Added block to admin detection tools when hash calculation is running.
* admin: Added more options to calculate date hashes faster.

### DIFF
--- a/ote/resources/public/language/en.edn
+++ b/ote/resources/public/language/en.edn
@@ -1293,7 +1293,7 @@ You can also draw the operating area or point to the map with drawing tools. You
   :limit-search-results "Limit search results"
   :participating-operator "Participates in providing of this service"
   :released-apis "Interfaces published in the service"
-  :no-description "Palvelulle ei ole annettu kuvausta."
+  :no-description "No description has been given to the service yet."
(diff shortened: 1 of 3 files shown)


ote/src/clj/ote/services/transit_changes.clj
```diff
@@ -106,37 +106,50 @@
...
 
-  (GET "/transit-changes/force-monthly-day-hash-calculation" {user :user}
+  ;; Calculate date-hashes for services using latest monthly package
+  (GET "/transit-changes/force-monthly-day-hash-calculation/:future" [future :as {user :user}]
     (when (authorization/admin? user)
       (http/transit-response
         {:status 200
-         :body (detection/calculate-monthly-date-hashes-for-packages db)})))
+         :body (detection/calculate-monthly-date-hashes-for-packages
+                 db (:user user) (if (= "true" future) true false))})))
```
mattiuusitalo: Doesn't (= "true" future) return true / false without the outer if?

rckopteri: Nice! yes it does.

PR #910, opened 2019-02-26
164237576 pre navigate event

philosotractor:
# Changed
* declare pre-navigate-event
* transit-visualization pre-navigate-event to clear view state

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -193,6 +193,14 @@
              {:on-success (tuck/send-async! ->LoadServiceChangesForDateResponse detection-date)})
   (init-view-state app))
 
+
+(define-event PreInitTransitVisualization []
+  {}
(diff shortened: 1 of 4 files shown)


philosotractor:
Will test this some more before merging. It occurred to me why wouldn't two consecutive tuck-events sent from controller do the same thing...maybe, maybe not.

philosotractor:
This doesn't do what we needed because 
`ote.app.routes/on-navigate` sets a `.setTimeout` and swaps the appstate, which triggers the view re-rendering. Which starts to draw the view using the app state data.
Then the events queued by the `.setTimeout`are triggered: `on-leave`and `on-navigate`, which manipulate appstate which triggers re-rendering of view using the new data.

This causes the view flickering and adding a new route change event like this doesn't fix the problem on FF especially. The desired solution has to update appstate before the swap. 
Will study some more...

philosotractor:
Closing this as there's a better alternative on my mind, there should be a generic improvement to the front-end design which fixes the issue everywhere.

PR #916, opened 2019-03-04
164047545 no route changes message

rckopteri:
# Added
* Added own style file for table styles.
   
# Changed
* table component:  Changed container element for no rows message from common help to plain white div.


ote/src/cljs/ote/style/table.cljs
```diff
@@ -0,0 +1,9 @@
+(ns ote.style.table
+  "Table styles."
+  (:require
```
mattiuusitalo: Is this the correct place for these :requires? I don't see references to these symbols in the file.

PR #923, opened 2019-03-11
Update clojurescript and clojure

erno:
Let's see how this fares on CircleCI - seems fine on local dev setup

### DIFF
--- a/ote/src/clj/ote/components/db.clj
+++ b/ote/src/clj/ote/components/db.clj
@@ -1,7 +1,6 @@
 (ns ote.components.db
   "Database connection pool component"
-  (:import (com.zaxxer.hikari HikariConfig HikariDataSource)
-           (org.postgresql.Driver))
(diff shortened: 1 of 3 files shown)


ote/src/clj/ote/components/db.clj
```diff
@@ -1,7 +1,6 @@
 (ns ote.components.db
   "Database connection pool component"
-  (:import (com.zaxxer.hikari HikariConfig HikariDataSource)
-           (org.postgresql.Driver))
+  (:import (com.zaxxer.hikari HikariConfig HikariDataSource))
```
philosotractor: Just wodering why `org.postgresql.Driver` is removed, it's not needed?

erno: This form of one-word import form didn't have any effect, it was actually forbidden syntax according to spec but older Clojure versions didn't flag it.

PR #926, opened 2019-03-12
164000883 detected changes count in calendar

rckopteri:
# Fixed
* transit-visualization page: Fixed detected changes count in calendar component
* detection.clj backend: Counts were not calculated for ending or new routes. Calculation added.

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -220,31 +220,52 @@
      ;; Add all received routes to shown map
      [:compare :show-route-lines] merge (zipmap route-line-names (repeat true)))))
 
+(defn compare-stop-differences [transit-visualization date1-trips date2-trips]
+  (if-let [first-common-stop (transit-changes/first-common-stop (concat date1-trips date2-trips))]
+          (let [first-common-stop
+                #(assoc %
+                   :first-common-stop first-common-stop
+                   :first-common-stop-time (transit-changes/time-for-stop % first-common-stop))
+                date1-trips (mapv first-common-stop date1-trips)
+                date2-trips (mapv first-common-stop date2-trips)
+                combined-trips (transit-changes/merge-by-closest-time
+                                 :first-common-stop-time
+                                 date1-trips date2-trips)]
+
+            (assoc-in transit-visualization [:compare :combined-trips]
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -220,31 +220,52 @@
...
-
-      ;; Can't find common stop
-      (assoc-in transit-visualization [:compare :combined-trips] nil))
-
-    ;; Both dates not fetched, don't try to calculate
-    (assoc-in transit-visualization [:compare :combined-trips] nil))))
+    (cond
+      (and (not (:route-trips-for-date1-loading? transit-visualization)) ;; both trips should be loaded
+           (not (:route-trips-for-date2-loading? transit-visualization))
+           (not (empty? date1-trips))
```
philosotractor: `(not (empty?` could be replaced using `not-empty` on this and below rows.

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -220,31 +220,52 @@
...
-               (mapv (fn [[l r]]
-                       [l r (transit-changes/trip-stop-differences l r)])
-                     combined-trips)))
-
-      ;; Can't find common stop
-      (assoc-in transit-visualization [:compare :combined-trips] nil))
-
-    ;; Both dates not fetched, don't try to calculate
-    (assoc-in transit-visualization [:compare :combined-trips] nil))))
+    (cond
```
mattiuusitalo: Both 1. and 2. condition are almost identical and test if loading is still ongoing. If so, skip anything in this condition. This could be a good case for nested ifs, actually, since then you wouldn't have these massive condition clauses interleaved with short bits of code. Something like
(if still-loading transit-visualization (if (not-empty? date2-trips.....

ote/src/cljs/ote/app/controller/transit_visualization.cljs
rckopteri: Changed

ote/src/cljs/ote/app/controller/transit_visualization.cljs
rckopteri: It is so much cleaner and shorter this way. Thanx, changed.

PR #930, opened 2019-03-14
164600197 hash calculation to new thread

rckopteri:
# Fixed
* Transit changes backend: Start hash calculations in different thread and return ok response
* Admin transit changes: do not block route hash id type list.
* Admin transit changes: Start change detection automatically when date hashes are calculated.

### DIFF
--- a/ote/src/clj/ote/services/transit_changes.clj
+++ b/ote/src/clj/ote/services/transit_changes.clj
@@ -1,22 +1,24 @@
 (ns ote.services.transit-changes
-  (:require [ote.components.service :refer [define-service-component]]
-            [ote.components.http :as http]
-            [compojure.core :refer [GET POST]]
+  (:require [compojure.core :refer [GET POST]]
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/services/transit_changes.clj
```diff
@@ -111,26 +113,35 @@
...
-                     (= scope "month")
-                     (detection/calculate-monthly-date-hashes-for-packages
-                       db (:user user) (= "true" future))
-                     (= scope "day")
-                     (detection/calculate-date-hashes-for-all-packages
-                       db (:user user) (= "true" future))
-                     (= scope "contract")
-                     (detection/calculate-date-hashes-for-contract-traffic
-                       db (:user user) (= "true" future)))})))
+         :body "OK"})))
```
philosotractor: If we want to be exact, result in this case should be `202 Accepted` indicating processing is started but not completed.

PR #933, opened 2019-03-15
152841059 spatial search v1

mattiuusitalo:
# Changed
New implementation for operation area search which is based on spatial joins instead of place names
   
# Datamodel

- PLACES view has been updated so SRID is set to 4326 for every geometry column. Otherwise spatial joins won't work.
- New tables spatial-relations-places and spatial-relations-custom-areas that are prefilled so using costly PostGIS functions for spatial querying is avoided.
- Added primary? column to "operation-area-facet" table so secondary operation areas are more easily filtered out of the search results


ote/src/clj/ote/services/service_search.clj
```diff
@@ -30,13 +30,37 @@
...
+  "Returns match quality of a search result to the operation-area it was searched against. Greater return values are worse matches.
+   Negative return value is an invalid match"
+  (if (pos? intersection) (/ difference intersection) -1))
+
+(defn sort-by-match-quality
+  "Sorts results by the match quality. Removes results with invalid match qualities"
+  [results spatial-diffs]
+  (let [quality-by-id (zipmap (map :id spatial-diffs) (map #(match-quality (:intersection %) (:difference %)) spatial-diffs))]
+    (->> results
+         (map #(assoc % :match-quality (get quality-by-id (::t-service/id %))))
```
rckopteri: No need to change this now, but it seems to me that this could be simpler.
If (match-quality) fn would take row as parameter and would add quality as map parameter when quality is positive and nil if negative. Then we would not need zip map and this map could be a keep fn, with call to match-quality.

ote/src/clj/ote/services/service_search.sql
```diff
@@ -125,3 +125,33 @@ SELECT eid."transport-service-id" as id
...
+-- Find services by operation area names
+SELECT oa."transport-service-id" as id
+  FROM "operation_area" oa,
+      (SELECT ST_Union(array_agg(pl.location)) as "location"
+         FROM places pl
+        WHERE pl.namefin IN (:operation-area)) as "sa",
+       "transport-service" ts
+ WHERE ts.published IS NOT NULL
+   AND oa."primary?" = true
+   AND ts.id = oa."transport-service-id"
```
rckopteri: Should service be public?

ote/test/clj/ote/format_test.clj
```diff
@@ -0,0 +1,12 @@
+(ns ote.format-test
+  (:require  [clojure.test :as t :refer [deftest is]]
+             [ote.format :as sut]))
+
+(deftest formatting-municipalities-with-postal-code
+  (is (= "Takalo 89660" (sut/postal-code-at-end "89660 Takalo"))
```
rckopteri: nice

ote/src/clj/ote/services/service_search.clj
mattiuusitalo: Currently I have a collection of results and a collection of spatial differences. How can I pick the correct spatial difference for a result without the map? I think this is where some of the idea of first get the keys, then get the values is beginning to break down a little bit. I definitely could change map -> filter to one keep call, that much I can understand.

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -273,8 +274,26 @@
 
 (def transport-types [:road :rail :sea :aviation])
 
+(defn- operation-area-searchbox [e! operation-area-filter-completions]
+  (let [suggestions (mapv (fn [s] {:text (format/postal-code-at-end s) :value s}) operation-area-filter-completions)]
+    {:name ::t-service/operation-area
+     :type :chip-input
+     :container-class "col-xs-12 col-sm-4 col-md-4"
+     :hint-text (tr [:service-search :operation-area-search-placeholder])
+     :filter (constantly true)
```
philosotractor: Was it checked if a plain `true` would work as well?

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -273,8 +274,26 @@
...
+     :hint-style {:top "20px"}
```
philosotractor: Wouldn't `rem` work better in case user changes the text size in browser (ctrl/cmd+ or -)

ote/src/cljs/ote/views/service_search.cljs
mattiuusitalo: filter property is a callback function in material ui's autocomplete component

https://v0.material-ui.com/#/components/auto-complete

ote/src/cljs/ote/app/controller/service_search.cljs
```diff
@@ -43,7 +45,7 @@
                        :as filters}]
   (merge
    (when-not (empty? oa)
-     {:operation_area (str/join "," (map :text oa))})
+     {:operation_area  (str/join "," (map :value oa))})
```
philosotractor: @mattiuusitalo Your editor seems to use different formatting rules as we have as convention in the project.
E.g. here there should be just one space between `:operation_area  (`
Similar diff can be seen in this file startoing around line 382

database/src/main/resources/db/migration/V1_141__set_strid_to_places_view.sql
```diff
@@ -0,0 +1,36 @@
+--- Rephrase view so that SRIDs are set to 4326.
```
philosotractor: Question, shoud `4326` be documented somewhere?

ote/src/cljs/ote/views/service_search.cljs
mattiuusitalo: Yes

ote/src/cljs/ote/app/controller/service_search.cljs
mattiuusitalo: The other diffs were poorly indented originally. The one highlighted here is definitely poor formatting.

ote/src/cljs/ote/app/controller/service_search.cljs
```diff
@@ -104,10 +106,8 @@
       ;; a service page.
       app
 
-      ;; Otherwise fetch facets and do a search
+      ;; Otherwise and do an initial search
```
jkorpelin: Unnecessary and?

ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -352,54 +371,34 @@
...
-          :open-on-focus? true}
+           {:id "sub-types"
+            :name ::t-service/sub-type
+            :label (tr [:service-search :type-search])
+            :type :chip-input
+            :container-class "col-xs-12 col-sm-4 col-md-4"
+            :full-width? true
+            :full-width-input? false
+            :suggestions-config {:text :text :value :text}
+            :suggestions (sub-types-to-list (::t-service/sub-type facets))
```
mattiuusitalo: Removing facets has broken autosuggestions here

PR #938, opened 2019-03-20
164600197 reset calculation status

rckopteri:
# Fixed
* Admin: Hash calculation status might get stuck in testing environment when deploying new package. Added button to reset latest calculation status.
* Fixed few private functions to remove warnings from figwheel.

# Added
* Admin: Added form to enable change detection for given service.

### DIFF
--- a/ote/src/clj/ote/services/transit_changes.clj
+++ b/ote/src/clj/ote/services/transit_changes.clj
@@ -112,6 +112,10 @@
        (when (authorization/admin? user)
          (http/transit-response (detection/hash-recalculations db))))
 
+  (GET "/transit-changes/reset-hash-calculation-status" {user :user :as request}
+    (when (authorization/admin? user)
+      (http/transit-response (detection/reset-last-hash-recalculations db))))
+
   ;; Calculate date-hashes. day/month/contract (all or only latest on every month or only for contract traffic) true/false (only to future or all days)
   (GET "/transit-changes/hash-calculation/:scope/:future" [scope is-future :as {user :user}]
     (when (authorization/admin? user)
@@ -158,6 +162,12 @@
   (POST "/transit-changes/force-detect" req
         (when (authorization/admin? (:user req))
           (gtfs-tasks/detect-new-changes-task db (time/now) true)
+          "OK"))
+
+  ;; Force detection change for all services
+  (GET "/transit-changes/detect-changes-for-given-service/:service-id" [service-id :as {user :user}]
+        (when (authorization/admin? user)
+          (gtfs-tasks/detect-new-changes-task db (time/now) true [(Long/parseLong service-id)])
           "OK"))
 
   ;; Force detection change for all services for given date
--- a/ote/src/clj/ote/services/transit_changes.clj
+++ b/ote/src/clj/ote/services/transit_changes.clj
@@ -1,5 +1,5 @@
 (ns ote.services.transit-changes
-  (:require [compojure.core :refer [GET POST]]
+  (:require [compojure.core :refer [GET POST DELETE]]
             [jeesql.core :refer [defqueries]]
             [clojure.string :as str]
             [clojure.set :as set]
@@ -112,7 +112,7 @@
        (when (authorization/admin? user)
          (http/transit-response (detection/hash-recalculations db))))
 
-  (GET "/transit-changes/reset-hash-calculation-status" {user :user :as request}
+  (DELETE "/transit-changes/hash-calculation" {user :user :as request}
     (when (authorization/admin? user)
       (http/transit-response (detection/reset-last-hash-recalculations db))))
 
@@ -158,26 +158,18 @@
     (when (authorization/admin? (:user req))
       (http/transit-response (services-with-route-hash-id db))))
 
-  ;; Force detection change for all services
-  (POST "/transit-changes/force-detect" req
+  ;; Force change detection for all services
+  (POST "/transit-changes/force-detect/" req
         (when (authorization/admin? (:user req))
           (gtfs-tasks/detect-new-changes-task db (time/now) true)
           "OK"))
 
-  ;; Force detection change for all services
-  (GET "/transit-changes/detect-changes-for-given-service/:service-id" [service-id :as {user :user}]
+  ;; Force change detection for single service
+  (POST "/transit-changes/force-detect/:service-id" {{:keys [service-id]} :params
+                                                     user :user}
         (when (authorization/admin? user)
           (gtfs-tasks/detect-new-changes-task db (time/now) true [(Long/parseLong service-id)])
           "OK"))
-
-  ;; Force detection change for all services for given date
-  (POST "/transit-changes/force-detect-date/:date"
-        {{:keys [date]} :params
-         user           :user
-         :as            req}
-    (when (authorization/admin? user)
-      (gtfs-tasks/detect-new-changes-task db (time/date-string->date-time date) true)
-      "OK"))
 
   ;; Delete row from gtfs_package to make this work. Don't know why, but it must be done.
   ;; Also change external-interface-description.gtfs-imported to past to make import work because we only import new packages.
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -41,6 +41,18 @@
     (when (pos-int? (count calculations))
        {:calculations calculations})))
 
+(defn reset-last-hash-recalculations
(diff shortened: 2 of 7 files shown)


ote/src/clj/ote/services/transit_changes.clj
```diff
@@ -112,6 +112,10 @@
        (when (authorization/admin? user)
          (http/transit-response (detection/hash-recalculations db))))
 
+  (GET "/transit-changes/reset-hash-calculation-status" {user :user :as request}
```
philosotractor: With same effort this could be made RESTful if taking following aproach:

```suggestion
  (DELETE "/transit-changes/hash-calculation/" {user :user :as request}
       (when (authorization/admin? user)
         (http/transit-response (detection/hash-recalculations db))))
```

ote/src/clj/ote/services/transit_changes.clj
```diff
@@ -160,6 +164,12 @@
           (gtfs-tasks/detect-new-changes-task db (time/now) true)
           "OK"))
 
+  ;; Force detection change for all services
+  (GET "/transit-changes/detect-changes-for-given-service/:service-id" [service-id :as {user :user}]
```
philosotractor: 1. This basically same as following existing endpoint:
```
;; Calculate date-hashes for given service-id and package-count
  (GET "/transit-changes/force-calculate-hashes/:service-id/:package-count" [service-id package-count :as {user :user}]
```

Would it bring benefit to try to follow RESTful principles here as well?
In which case this would be nd "overload" to force-calculate-hashes, something like:
``` 
 ;; Calculate date-hashes for given service-id and package-count
  (GET "/transit-changes/force-calculate-hashes/:service-id" [service-id :as {user :user}]
    (when (authorization/admin? user)
      (do
        ;; Calculate date hashes in other thread
        (future
          (detection/calculate-package-hashes-for-service db (Long/parseLong service-id) (Long/parseLong package-count) (:user user))
          ;; Detect changes for service
          (gtfs-tasks/detect-new-changes-task db (time/now) true [(Long/parseLong service-id)]))
"OK")))
```

2. Note that `force-calculate-hashes` uses `future` threading to perform the same task while this doesn't. Probably this should - which gets us to comment part 1. here :)

ote/src/clj/ote/services/transit_changes.clj
philosotractor: Had a copy paste mistake, updated the snippet now :)

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -607,7 +619,7 @@
 
 (defn- max-date-within-90-days? [{max-date :max-date}]
```
philosotractor: Function name will be misleading after this, I suggest renaming it since it's referenced in only one place in the code
```
(defn- max-date-within-evaluation-window? [{max-date :max-date}]
```

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -607,7 +619,7 @@
...
   (and max-date
-       (.isBefore (.toLocalDate max-date) (.plusDays (java.time.LocalDate/now) 90))
+       (.isBefore (.toLocalDate max-date) (.plusDays (java.time.LocalDate/now) 180))
```
philosotractor: Function name will be misleading after this, I suggest renaming it since it's referenced in only one place in the code. E.g.
```
(defn- max-date-within-evaluation-window? [{max-date :max-date}]
```

ote/src/cljs/ote/views/admin/detected_changes.cljs
```diff
@@ -35,7 +35,15 @@
                               (.preventDefault %)
                               (e! (admin-transit-changes/->LoadHashRecalculations)))
                  :icon (ic/content-filter-list)})
-      [:span "Tarkista laskennan eteneminen"]]]))
+      [:span "Tarkista laskennan eteneminen"]]
+     [:a (merge (stylefy/use-style button-styles/negative-button)
```
philosotractor: A bit nitpicking: 
There generic nap UI button components are defined in `ote/ui/buttons`, where `/delete`matches to this need style-wise. Might be good to use those whenever possible?

ote/src/cljs/ote/views/admin/detected_changes.cljs
```diff
@@ -97,19 +105,41 @@
        [:span "Pakota yhden lataamattoman pakettin lataus ulkoisesta osoitteesta"]]]]
 
     [:h2 "Muutostunnistuksen käynnistys"]
+    [:div
+     [:div (stylefy/use-style style-admin/detection-button-container)
+      [:div (stylefy/use-style style-admin/detection-info-text)
+       [:span "Pakota muutostunnistus kaikkille palveluille. Tämä vie noin 3 minuuttia."]]
+      [:div {:style {:flex 2}}
+       [:a (merge (stylefy/use-style button-styles/primary-button)
```
philosotractor: Is there a specific reason for using `[:a` instead of a `button`? Buttons should be used for actions, links for navigation.
Could `save` in `ote/ui/buttons.cljs` be used, if not then define an overload button there?

ote/src/cljs/ote/views/admin/detected_changes.cljs
```diff
@@ -97,19 +105,41 @@
...
+       [ui/text-field
+        {:id                  "jaakko-mies"
+         :name                "jes"
+         :floating-label-text "Palvelun id"
+         :value               (get-in app-state [:admin :transit-changes :single-detection-service-id])
+         ;:type    :string
+         :on-change           #(do
+                                 (.preventDefault %)
+                                 (e! (admin-transit-changes/->SetSingleDetectionServiceId %2)))}]
+       [:a (merge (stylefy/use-style button-styles/primary-button)
```
philosotractor: Is there a specific reason for using `[:a` instead of a `button`? Buttons should be used for actions, links for navigation.
Could `save` in `ote/ui/buttons.cljs` be used, if not then define an overload button there?

ote/src/cljs/ote/views/admin/detected_changes.cljs
```diff
@@ -97,19 +105,41 @@
...
-                               (.preventDefault %)
-                               (e! (admin-transit-changes/->ForceDetectTransitChanges)))
-                  :icon (ic/content-filter-list)})
-       [:span "Käynnistä muutostunnitus"]]]]
+     [:div (stylefy/use-style style-admin/detection-button-container)
+      [:div {:style {:flex 2}} "Käynnistä muutostunnistus vain yhdelle palvelulle. Anna kenttään palvelun id.
+      Löydät palvelun id:n omat palvelutiedot sivun kautta tai muutostunnistuksen visualisointisivun url:stä."]
+      [:div {:style {:flex 2}}
+       [ui/text-field
+        {:id                  "jaakko-mies"
```
philosotractor: Should id be renamed? :)

ote/src/clj/ote/services/transit_changes.clj
philosotractor: Okay well this initial comment was written after only a quick look so the proposed implementation is not fully kosher and well thought, but pleaase consider the API point of view :)

philosotractor:
LGTM as-is, though made a few comments on how the API could be structured in a bit different way.
Please rename the

philosotractor:
LGTM as-is, though made a few comments on how the API could be structured in a bit different way.
Renaming the max date function would be nice, those kind of things are easy to forget.

ote/src/clj/ote/services/transit_changes.clj
rckopteri: Changed

ote/src/clj/ote/services/transit_changes.clj
rckopteri: Force-calculate-hashes is different than detect-changes.
So it is better to refactor those API endpoints to make this more clear.

ote/src/clj/ote/transit_changes/detection.clj
rckopteri: Good idea!

ote/src/cljs/ote/views/admin/detected_changes.cljs
rckopteri: Nope. Changed

ote/src/cljs/ote/views/admin/detected_changes.cljs
rckopteri: What? I think that this is just what we need.

PR #940, opened 2019-03-21
164756947 route summary

rckopteri:
# Added
* Transit-visualization page: Added summary of changes to all route rows
* Transit-changes page: Added colors to change icons at the top of the page
* Transit-visualization page: Added new route-ending icon.
* Transit visualization page: Added more colors to calendar to visualize traffic

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -19,7 +19,7 @@
   (or (:route-hash-id route) (str (:route-short-name route) "-" (:route-long-name route) "-" (:trip-headsign route))))
 
 (def hash-colors
-  ["#E1F4FD" "#DDF1D2" "#FFF7CE" "#E0B6F3" "#A4C9EB" "#FBDEC4" "#FFAD9A" "#BDECB7" "#BAA2D1" "#F89DC0" "#97D5E8" "#E1C7B0" "#9CCDDC" "#c82565"]
+  ["#E1F4FD" "#DDF1D2" "#FFF7CE" "#E0B6F3" "#A4C9EB" "#FBDEC4" "#FFAD9A" "#BDECB7" "#BAA2D1" "#F89DC0" "#97D5E8" "#E1C7B0" "#9CCDDC" "#c56c8f" "#b9cf84" "#a4515b" "#23dbe1"]
   #_["#52ef99" "#c82565" "#8fec2f" "#8033cb" "#5c922f" "#fe74fe" "#02531d"
    "#ec8fb5" "#23dbe1" "#a4515b" "#169294" "#fd5925" "#3d4e92" "#f4d403"
    "#66a1e5" "#d07d09" "#9382e9" "#b9cf84" "#544437" "#f2cdb9"])
--- a/ote/src/cljs/ote/views/transit_visualization.cljs
+++ b/ote/src/cljs/ote/views/transit_visualization.cljs
@@ -2,7 +2,6 @@
   "Visualization of transit data (GTFS)."
   (:require [reagent.core :as r]
             [cljs-react-material-ui.icons :as ic]
-            [ote.ui.icons :as ote-icons]
             [ote.ui.icon_labeled :as icon-l]
             [stylefy.core :as stylefy]
             [ote.style.transit-changes :as style]
@@ -313,7 +312,7 @@
         all-route-changes (filter
                             (fn [c]
                               (let [x (select-keys c selected-change-keys)]
-                                (if (= (:route-hash-id x) (:route-hash-id single-change)) true false)))
+                                (= (:route-hash-id x) (:route-hash-id single-change))))
                             all-changes)
         merged-changes (apply merge-with + all-route-changes)]
     merged-changes))
@@ -381,22 +380,22 @@
         :format (fn [{change-type :change-type :as route-changes}]
                   (case change-type
                     :no-traffic
-                    [tv-change-icons/labeled-icon
+                    [icon-l/icon-labeled
                      [ic/av-not-interested]
                      "Tauko liikennöinnissä"]
 
                     :added
-                    [tv-change-icons/labeled-icon
+                    [icon-l/icon-labeled
                      [ic/content-add-box {:color style/add-color}]
                      "Uusi reitti"]
 
                     :removed
-                    [tv-change-icons/labeled-icon
+                    [icon-l/icon-labeled
                      [ic/content-remove-circle-outline {:color style/remove-color}]
                      "Mahdollisesti päättyvä reitti"]
 
                     :no-change
-                    [tv-change-icons/labeled-icon
+                    [icon-l/icon-labeled
                      [ic/navigation-check]
                      "Ei muutoksia"]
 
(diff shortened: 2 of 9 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -19,7 +19,7 @@
   (or (:route-hash-id route) (str (:route-short-name route) "-" (:route-long-name route) "-" (:trip-headsign route))))
 
 (def hash-colors
-  ["#E1F4FD" "#DDF1D2" "#FFF7CE" "#E0B6F3" "#A4C9EB" "#FBDEC4" "#FFAD9A" "#BDECB7" "#BAA2D1" "#F89DC0" "#97D5E8" "#E1C7B0" "#9CCDDC" "#c82565"]
+  ["#E1F4FD" "#DDF1D2" "#FFF7CE" "#E0B6F3" "#A4C9EB" "#FBDEC4" "#FFAD9A" "#BDECB7" "#BAA2D1" "#F89DC0" "#97D5E8" "#E1C7B0" "#9CCDDC" "#c56c8f" "#b9cf84" "#a4515b" "#23dbe1"]
```
philosotractor: In the future would be good to have def's for these in stye file with a semantic name. `hash-colors` vector should then refer to those symbols.

ote/src/cljs/ote/views/transit_visualization/change_utilities.cljs
```diff
@@ -35,9 +35,10 @@
     [:b "Taulukon ikonien selitteet"]]
    [:div (stylefy/use-style style/transit-changes-icon-legend-row-container)
     (doall
-      (for [[icon label] [[ote-icons/outline-add-box " Uusia vuoroja"]
-                          [ote-icons/outline-indeterminate-checkbox " Poistuvia vuoroja"]
-                          [ic/action-timeline " Pysäkkimuutoksia per vuoro"]
-                          [ic/action-query-builder " Aikataulumuutoksia per vuoro"]]]
+      (for [[icon color label] [[ic/content-remove-circle-outline {:color style/remove-color} "Mahdollisesti päättyvä reitti"]
```
philosotractor: Old code I know, but still: a bit 90s to have marginal in localization string as a space character.:)
Alternatively could be done using styling or by prepending the space " " to the string here and not in the language string itself. The minor value there is it enables potential re-using of the string in some other view place i a need arises :)

ote/src/cljs/ote/views/transit_visualization/change_utilities.cljs
```diff
@@ -35,9 +35,10 @@
...
+                          [ote-icons/outline-add-box {} " Uusia vuoroja"]
```
philosotractor: Old code I know, but still: a bit 90s to have marginal in localization string as a space character.:)
Alternatively could be done using styling or by prepending the space " " to the string here and not in the language string itself. The minor value there is it enables potential re-using of the string in some other view place i a need arises :)

ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
```diff
@@ -39,23 +40,33 @@
 
 
 ;; Ui
+(defn labeled-icon
```
philosotractor: Apparently there was a problem with ´ote.ui.icon_labeled´ and it couldn't be used or developed further?

ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -600,8 +659,10 @@
          {:icon (if open?
                   [ic/navigation-expand-less]
                   [ic/navigation-expand-more])
-          :on-click #(do (.preventDefault %)
-                         (e! (tv/->ToggleSection :gtfs-package-info)))
+          :on-click (fn [^SyntheticMouseEvent event]
```
philosotractor: 👍

ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -324,25 +346,62 @@
...
+                    (:count row)))}
+       {:name "Aikaa 1. muutokseen"
+        :width "15%"
+        :read :different-week-date
+        :format (fn [different-week-date]
+                  (if-not different-week-date
+                    [tv-change-icons/labeled-icon [ic/navigation-check] "Ei muutoksia"]
+                    [:span
+                     (str (time/days-until different-week-date) " pv")
+                     [:span (stylefy/use-style {:margin-left "5px"
```
philosotractor: Did you notice there's `icon-labeled-icon` intended to be used with e.g. `ote.ui.icon_labeled`. That gives a relative margin between the icon and label.
What's needed here more is the `:color` overload, which could be done by extending `icon-labeled-icon` using a color from style definitions we have. The colours there in turn come from the UI specification, which give us the semantic name and color code.

ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -297,7 +301,24 @@
...
+  [single-change all-changes]
+  (let [selected-change-keys #{:removed-trips :trip-stop-sequence-changes-lower
+                               :trip-stop-sequence-changes-upper :route-hash-id
+                               :trip-stop-time-changes-lower :trip-stop-time-changes-upper :change-type :added-trips}
+        all-changes (map #(select-keys % selected-change-keys) all-changes)
+        single-change (select-keys single-change selected-change-keys)
+        all-route-changes (filter
+                            (fn [c]
+                              (let [x (select-keys c selected-change-keys)]
+                                (if (= (:route-hash-id x) (:route-hash-id single-change)) true false)))
```
philosotractor: Alternative
```suggestion
                                (= (:route-hash-id x) (:route-hash-id single-change)) true false))
```

ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -297,7 +301,24 @@
...
+  need to merge-with them together."
+  [single-change all-changes]
+  (let [selected-change-keys #{:removed-trips :trip-stop-sequence-changes-lower
+                               :trip-stop-sequence-changes-upper :route-hash-id
+                               :trip-stop-time-changes-lower :trip-stop-time-changes-upper :change-type :added-trips}
+        all-changes (map #(select-keys % selected-change-keys) all-changes)
+        single-change (select-keys single-change selected-change-keys)
+        all-route-changes (filter
+                            (fn [c]
+                              (let [x (select-keys c selected-change-keys)]
```
philosotractor: Is the `let` unnecessary if function only examines two explicitly defined keys in the following condition?

ote/src/cljs/ote/app/controller/transit_visualization.cljs
rckopteri: We need correct colors from Jussi. Eight of these are my selections and there are not good.
Would be appropriate to move these to correct place when we get correct colors.

ote/src/cljs/ote/views/transit_visualization.cljs
rckopteri: These are the same old codes that I removed when multiple changes were introduced in ui. 
We have bit of an issue with these icon - label thingies in this view. There isn't enough space to all of this information. So I leave it as it is because it works kind of.

ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
rckopteri: ah. My bad. This "labeled-icon" was a copy&paste of ote.ui.icon_labeled "icon-labeled" function.
And it was used in transit_visualization view. So I only moved it to change_icon file.

So I now removed that copy&paste and take icon-labeled into use.

ote/src/cljs/ote/views/transit_visualization/change_utilities.cljs
rckopteri: :)

PR #944, opened 2019-03-28
163148396 Clean up wrong change-dates from db

rckopteri:
# Added
* Add tests for 25 weeks and 35 weeks to ensure that 30 week detection window works
* I had to refactor detection test codes. I didn't change them. I only added two more.

### DIFF
--- a/database/src/main/resources/db/migration/V1_142__remove_old_change_dates.sql
+++ b/database/src/main/resources/db/migration/V1_142__remove_old_change_dates.sql
@@ -1,0 +1,5 @@
+-- Remove gtfs-transit-changes.change-date dates from database to prevent unwanted change-detections
+-- In local and test environments there might be correct change-date's so clean up only those that are within the next week (7 days)
+UPDATE "gtfs-transit-changes" SET "change-date" = null
+ WHERE "change-date" >= CURRENT_DATE
+   AND "change-date" <= (CURRENT_DATE + '7 day'::interval) ;
--- a/ote/test/clj/ote/transit_changes/detection_test_utilities.clj
+++ b/ote/test/clj/ote/transit_changes/detection_test_utilities.clj
@@ -1,0 +1,27 @@
+(ns ote.transit-changes.detection-test-utilities
+	"Helper methods for detection tests."
+	(:require [clojure.test :as t :refer [deftest testing is]]
+						[ote.time :as time]))
+
+(defn to-local-date [year month day]
+	(java.time.LocalDate/of year month day))
+
+(def select-keys-detect-changes-for-all-routes [:different-week
+																								:no-traffic-start-date
+																								:no-traffic-end-date
+																								:no-traffic-run
+																								:no-traffic-change
+																								:route-key
+																								:starting-week])
+(def route-name "Raimola")
+(def route-name-2 "Esala")
+
+(defn weeks
+	"Give first day of week (monday) as a starting-from."
+	[starting-from route-maps]
+	(vec (map-indexed
+				 (fn [i routes]
+					 {:beginning-of-week (.plusDays starting-from (* i 7))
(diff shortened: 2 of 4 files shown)


database/src/main/resources/db/migration/V1_142__remove_old_change_dates.sql
```diff
@@ -0,0 +1,5 @@
+-- Remove gtfs-transit-changes.change-date dates from database to prevent unwanted change-detections
```
philosotractor: Doesn't this affect production as well? Current week's change dates would be lost, and re-created...when?

ote/test/clj/ote/transit_changes/detection_test_utilities.clj
```diff
@@ -0,0 +1,27 @@
+(ns ote.transit-changes.detection-test-utilities
```
philosotractor: Could code formatting be fixed, some of fields are far away on the right in the file.

ote/test/clj/ote/transit_changes/detection_window_test.clj
```diff
@@ -0,0 +1,55 @@
...
+(deftest test-change-at-week-33
+	(let [result (-> data-change-at-week-33
+									 detection/changes-by-week->changes-by-route
+									 detection/detect-changes-for-all-routes)]
+		(is (= {:route-key test-utilities/route-name
+						:starting-week {:beginning-of-week (test-utilities/to-local-date 2019 1 14)
+														:end-of-week (test-utilities/to-local-date 2019 1 20)}} ;; HOX! No different week, no change
+					 (-> result
+							 first
+							 (select-keys [:no-traffic-start-date :no-traffic-end-date :route-key :no-traffic-change
```
philosotractor: select-keys-detect-changes-for-all-routes

ote/test/clj/ote/transit_changes/detection_window_test.clj
```diff
@@ -0,0 +1,55 @@
...
+									 detection/changes-by-week->changes-by-route
+									 detection/detect-changes-for-all-routes)]
+		(is (= {:route-key test-utilities/route-name
+						:starting-week {:beginning-of-week (test-utilities/to-local-date 2019 1 14)
+														:end-of-week (test-utilities/to-local-date 2019 1 20)}
+						:different-week {:beginning-of-week (test-utilities/to-local-date 2019 6 17)
+														 :end-of-week (test-utilities/to-local-date 2019 6 23)}}
+					 (-> result
+							 first
+							 (select-keys [:no-traffic-start-date :no-traffic-end-date :route-key :no-traffic-change
```
philosotractor: select-keys-detect-changes-for-all-routes

database/src/main/resources/db/migration/V1_142__remove_old_change_dates.sql
rckopteri: I assume that we will get this pr to production in the next week or two. Next correct detection cycle starts about 4 weeks from now. So if there is change-days that are in the next 7 days, they are generated with wrong code.

ote/test/clj/ote/transit_changes/detection_window_test.clj
```diff
@@ -0,0 +1,55 @@
+(ns ote.transit-changes.detection-window-test
```
philosotractor: "design for testability" i.e. should the tested functions take the detection window value as an argument.
That way tests could test using different values...in the other hand would that bring any value?

ote/test/clj/ote/transit_changes/detection_window_test.clj
rckopteri: fixed

ote/test/clj/ote/transit_changes/detection_window_test.clj
rckopteri: fixed

PR #947, opened 2019-03-29
164941250 transit visualization ui tuning

rckopteri:
# Fixed
* Transit visualization: Add date of week to calendar change list
* calendar component: Fix month name for February when current day is greater than day count in February (29,30,31)
* All tables: Move header labels and column values to start at the same spot
* All tables: Fix long text values in tables by changing pre-wrap to pre-line
* transit-changes: Move change icons more to left

### DIFF
--- a/ote/src/cljc/ote/time.cljc
+++ b/ote/src/cljc/ote/time.cljc
@@ -163,6 +163,12 @@
   #?(:clj (java.time.LocalDate/of year month date)
      :cljs (goog.date.Date. year (dec month) date)))
 
+;; Change js-date (.js/date) to google datetime
+#?(:cljs
+   (defn js-date-to-goog-date [d]
+     (let [fields (date-fields d)]
+       (goog.date.Date. (:ote.time/year fields) (dec (:ote.time/month fields)) (:ote.time/date fields)))))
+
 (defn year [dt]
   (::year (date-fields dt)))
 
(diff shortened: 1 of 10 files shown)


ote/src/cljc/ote/time.cljc
```diff
@@ -163,6 +163,12 @@
   #?(:clj (java.time.LocalDate/of year month date)
      :cljs (goog.date.Date. year (dec month) date)))
 
+;; Change js-date (.js/date) to google datetime
+#?(:cljs
+   (defn js-date-to-goog-date [d]
```
jkorpelin: Should this be named `js-date->goog-date`?

ote/src/cljs/ote/views/transit_changes.cljs
```diff
@@ -61,21 +61,21 @@
 
 (defn change-icons [{:keys [added-routes removed-routes changed-routes no-traffic-routes]}]
   [:div.transit-change-icons (stylefy/use-style style/transit-changes-icon-row-container)
-   [:div (use-style (merge style/transit-changes-legend-icon {:width "25%"}))
+   [:div {:style {:width "25%"}}
```
jkorpelin: When `use-style` is used, stylefy creates a class for the style so it might be a bit more performant in some cases. Probably not a factor here since the in-line style is so short.

PR #948, opened 2019-03-29
164267877 services all info

jkorpelin:
# Added
* New service view
   
# Fixed
* Some paths having invalid authentication requirements.


ote/resources/public/language/en.edn
```diff
@@ -1349,6 +1352,38 @@ You can also draw the operating area or point to the map with drawing tools. You
   :check-email-for-link "Your request for changing password has been received. A link for resetting your password has now been sent to your email address."
   :logged-out "You have been logged out."}
 
+ :service-viewer
+ {:open-in-geojson "Avaa tiedot GeoJSON-muodossa"
```
rckopteri: Most of these or even maybe all are translated in viewer section above

ote/src/clj/ote/authorization.clj
```diff
@@ -41,6 +41,12 @@
       :else
       (body-fn))))
 
+(defn is-author?
+  [db user transport-operator-id]
+  (let [allowed-operators (user-transport-operators db user)]
+    (println "allowed operators: " allowed-operators)
```
rckopteri: remove println

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -0,0 +1,464 @@
...
+                                                              :top "6px"
+                                                              :padding-right "5px"
+                                                              :color style-base/link-color}]
+                                     (tr [:service-search :back-link])]]
+   [:h1 service-name]
+   [link-with-icon open-in-new-icon (str "/export/geojson/" o-id "/" s-id) (tr [:service-viewer :open-in-geojson])]])
+
+(defn- operator-info
+  [title operator]
+  (let [address (::t-operator/visiting-address operator)]
```
rckopteri: We should only show company name, www address and business id in this view. It is confusing to show different data for different users. It will cause problems.

ote/src/cljs/ote/ui/link_icon.cljs
```diff
@@ -0,0 +1,16 @@
...
+  (:require [stylefy.core :as stylefy]
+            [ote.style.base :as style-base]))
+
+
+(defn link-with-icon
+  [icon url link-text on-click]
+  [:div
+   [:a (merge {:href url
+               :style {:margin-right "2rem"}
+               :id "edit-transport-operator-btn"}
```
rckopteri: We can't know where this button is used so the id can't be named as transport operator

ote/src/cljs/ote/views/service_viewer.cljs
jkorpelin: This wasn't shown anywhere, but I removed the let.

ote/resources/public/language/en.edn
jkorpelin: They were slightly differently worded so I thought it was just safer to add new ones.

ote/src/cljs/ote/app/controller/transport_operator.cljs
```diff
@@ -521,5 +521,4 @@
 (defmethod routes/on-navigate-event :transport-operator [{id :params}]
   (if (some? (:id id))
     (->EditTransportOperator (:id id))
-    (->VerifyCreateState)
-    ))
+    (->VerifyCreateState)))
```
philosotractor: 👍

ote/src/cljs/ote/app/controller/service_viewer.cljs
```diff
@@ -0,0 +1,100 @@
...
+                                                    :id (::t-service/id area)
+                                                    :lang (::t-service/lang (first (::t-service/description area)))}]
+
+                                (if (::t-service/primary? area)
+                                  (assoc new-coll :primary (conj (:primary new-coll) formatted-area))
+                                  (assoc new-coll :secondary (conj (:secondary new-coll) formatted-area)))))
+                            {}
+                            areas)
+                    accessibility-data (let [pt (::t-service/passenger-transportation result)
+                                             guaranteed-vehicle-acc (set (::t-service/guaranteed-vehicle-accessibility pt))
```
philosotractor: Is the `set` for following symbol definitions?

ote/src/clj/ote/services/transport.clj
```diff
@@ -91,9 +102,10 @@
                                [::t-service/external-interfaces
                                 (specql/columns ::t-service/external-interface-description)])
                          {::t-service/id id}))]
-    (when ts
-      (assoc ts ::t-service/operation-area
-             (places/fetch-transport-service-operation-area db id)))))
+    (if ts
+      (http/transit-response (assoc ts ::t-service/operation-area
+                                       (places/fetch-transport-service-operation-area db id)))
+      {:status 404})))
```
philosotractor: 👍

ote/src/clj/ote/services/transport.clj
```diff
@@ -37,6 +37,17 @@
   #{::t-operator/id ::t-operator/business-id ::t-operator/email
     ::t-operator/name})
 
+(def transport-service-personal-columns
+  #{::t-service/contact-phone
+    ::t-service/contact-gsm
+    ::t-service/contact-email})
+
+(def transport-operator-personal-columns
```
philosotractor: `transport-operator-personal-columns` doesn't seem to be used anywhere

ote/src/clj/ote/services/transport.clj
```diff
@@ -471,24 +483,51 @@
       (log/info (str "Delete association failed with service: " service-id " and operator: " operator-id))
       {:status 404 :body "Association not found"})))
 
+(defn public-data-transport-service
+  "Get single transport service by id"
+  [db id]
+  (let [ts (first (fetch db ::t-service/transport-service
+                         (apply disj
```
philosotractor: I'm wondering what are the benefits between 
a) just defining the allowed columns
vs. 
b) taking "all" columns and disjoinin the "not-allowed" ones like done here.

In the first case it's a bit more quaranteed that no unwanted column gets to the result set accidentally because one has to add them explicitly, 
in this case any new column gets included by default.

Which ever design path should be taken? If b) is taken, then this solution looks good.

ote/src/clj/ote/authorization.clj
```diff
@@ -41,6 +41,11 @@
       :else
       (body-fn))))
 
+(defn is-author?
```
philosotractor: A matter of flavour, but should name be something like `user-in-group?`

ote/src/clj/ote/services/transport.clj
jkorpelin: I used this approach because I tought it was just temporary that we couldn't return everything.

ote/src/clj/ote/authorization.clj
rckopteri: I think that is-author? is really good option here. It separates admin and user level.

PR #949, opened 2019-04-01
163148387 state machine v3

rckopteri:
# Added
* detected-change-history table to store all unique changes
* transit-visualization page: correct day for detected date
* pre-notice email: Changes are sent only once.


ote/test/clj/ote/tasks/pre_notices_test.clj
```diff
@@ -35,6 +35,19 @@
 (defn- email-content []
   (get-in @outbox [0 :body 0 :content]))
 
+(defn- clean-up-db [db]
```
philosotractor: Shouldn't the package id be `def`'d in the test, and the test should pass the `def` as an argument to all the db CRUD functions?

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -710,6 +708,60 @@
 
 (spec/fdef transform-route-change
            :args (spec/cat :all-routes vector? :route-change ::service-route-change-map :route-changes-all ::detected-route-changes-for-services-coll))
+
+(defn- route-change-type [max-date-in-past? added? removed? changed? no-traffic? starting-week-date different-week-date
```
philosotractor: Indentation seems to be a bit off in this diff for this function body.

PR #952, opened 2019-04-04
161871710 transit visualization map section ui tuning

rckopteri:
# Changed
* Added different day indicators to map section
* Removed map from section header and moved it to section body
* Tuned a bit icon element widths

### DIFF
--- a/ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
+++ b/ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
@@ -47,7 +47,7 @@
             trip-stop-time-changes-lower trip-stop-time-changes-upper different-week-date] :as diff}
     with-labels?]
    [:div (stylefy/use-style (style-base/flex-container "row"))
-    [:div {:style {:width "15%"}}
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
```diff
@@ -47,27 +47,27 @@
...
        (when with-labels? " lisättyä vuoroa")]]]
 
-    [:div {:style {:width "15%"}}
+    [:div {:style {:width "20%"}}
      [icon-l/icon-labeled
       [ote-icons/outline-indeterminate-checkbox {:color (if (= 0 removed-trips)
                                                           style/no-change-color
                                                           style/remove-color)}]
-      [:span (or removed-trips (:gtfs/removed-trips diff))
+      [:span (or removed-trips (:gtfs/removed-trips diff) 0)
```
philosotractor: If a default `0` is added here, it should be added for `added-trips` as well. Same problem is there.
Though as we know, there's a backend fix for both now in master.

PR #955, opened 2019-04-04
Fix mime types for Mac OS X

rckopteri:
# Fixed
* Backup plan for mac os x machines with mime types


ote/src/clj/ote/services/pre_notices/attachments.clj
```diff
@@ -17,16 +17,23 @@
...
+  (:import (java.nio.file Files)
+           (java.net URLConnection)))
 
 (def allowed-mime-types #{"application/pdf" "image/jpeg" "image/png"})
 
 (defn- generate-file-key [id filename]
   (str id "_" filename))
 
-(defn validate-file [{:keys [tempfile]}]
+(defn validate-file [{:keys [tempfile filename]}]
```
philosotractor: PR is of course fine, but just thinking out loud that 
Following snippet is shorter because temporary symbols were referred only once. 
Also from performance point of view PR version relies on lazy evaluation, if that wasn't the case following would be more efficient.
What comes to readability and maintainability, it is a matter of debate :)

```
        ;; In Mac os x mime types are not handled very well in java 1.8. So we use this as a backup.
        content-mime (or (Files/probeContentType (.toPath tempfile))
                                      (URLConnection/guessContentTypeFromName filename))
```

philosotractor: Previous comment had a small thought glitch in the matrix: proposed code _is_ more efficient because in clj only map/for are lazy. So PR's version would fetch data for the temp sumbols even when if it's not used - the simplified proposal in previous comment wouldn't. :)

rckopteri: Nice!

PR #960, opened 2019-04-08
Update login credentials error message

rckopteri:
# Changed
* Update credentials error message


philosotractor:
Should we remove the unrelated commits from this branch & PR?

rckopteri:
I created another pr for those admin contract traffic changes and removed all commits related to that from this pr

PR #961, opened 2019-04-08
165113720 transit-visualization: hide past transit changes

philosotractor:
# Changed
* cljc: now-iso-date-str for getting current ISO format date in as string
* controller: transit-visualisation hide changes in past

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -101,7 +101,7 @@
         only-changes (sort-by :different-week-date (filterv :change-date changes))
 
         ;; Group by only-changes by route-hash-id
-        grouped-changes (group-by #(:route-hash-id %) only-changes)
+        grouped-changes (group-by :route-hash-id only-changes)
         ;; Take first from every vector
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -183,31 +183,35 @@
...
-    :service-changes-for-date-loading? false
-    :service-info (:service-info response)
-    :changes-all (sort-by :different-week-date < (:route-changes response))
-    :changes-route-no-change (sorted-route-changes true (future-changes detection-date (:route-changes response)))
-    :changes-route-filtered (sorted-route-changes false (future-changes detection-date (:route-changes response)))
-    :gtfs-package-info (:gtfs-package-info response)
-    :route-hash-id-type (:route-hash-id-type response)))
-
-(defn- init-view-state [app]
+              {:path [:transit-visualization]}
```
rckopteri: we have treated define-event macro as function so that next line indention is 2 spaces

philosotractor: Fix'd

PR #966, opened 2019-04-09
152841059 spatial search rank

mattiuusitalo:
# Fixed
*    Corrected spatial search result ranking when paging is used. Before this fix, spatial search results were ranked only withing the queried page.

### DIFF
--- a/ote/src/clj/ote/services/service_search.clj
+++ b/ote/src/clj/ote/services/service_search.clj
@@ -227,19 +227,20 @@ Negative return value is an invalid match"
               (latest-service-ids db)
               ;; Combine with intersection (AND)
               (apply set/intersection
-                     (remove nil? result-id-sets)))
-        results (if operation-area
+                     (remove nil? result-id-sets)))]
+    (-> (if operation-area
                   (transport-services-in-operation-area db ids operation-area offset limit)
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/services/service_search.clj
```diff
@@ -155,6 +151,83 @@ Negative return value is an invalid match"
...
+(defn- pare-results
+  "Removes information not intended for the frontend from `search-results`
+  Retains `searched-operators` in the search results"
+  [search-results searched-operators]
+  (->> search-results
+       (mapv without-import-errors)
+       (mapv (filtering-operators searched-operators))
+       (mapv without-personal-info)))
+
 (defn search [db {:keys [operation-area sub-type data-content transport-type text operators offset limit]
```
rckopteri: Just wondering that when we have more than 10 000 services in production this way of searching services might bite us. Maybe not the thing that we need to change right now, but this is far from optimal.

mattiuusitalo: The functionality here hasn't changed. I extracted this stuff into methods in an attempt to clarify what is being done. I suppose it could be possible to filter this stuff out in the queries themselves.

PR #967, opened 2019-04-10
Try updated cypress image

erno:


erno:
This Pr fulfilled its purpose, updated image pushed to tag latest

PR #971, opened 2019-04-12
165270343 & 165274555 service view text changes

jkorpelin:
# Added
* One more row of data to service-view
   
# Fixed
* Wording


ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -269,9 +269,9 @@
          [info-sections-2-cols (string/upper-case title)
           [:div
            [common-ui/information-row-with-option (tr [:service-search :homepage]) (when url [common-ui/linkify
-                                                                                           url
-                                                                                           url
-                                                                                           {:target "_blank"}]) false]
+                                                                                              url
```
rckopteri: Would be easier to read if parameters were on own lines starting from: (tr [:service-search :homepage])
Something like this:
[common-ui/information-row-with-option 
  (tr [:service-search :homepage]) 
  (when url  [common-ui/linkify url {:target "_blank"}]) 
  false]

PR #972, opened 2019-04-12
165304032 admin popup

jkorpelin:
# Fixed
* Admin popup
* timestamp when interface error

### DIFF
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -602,11 +602,11 @@
   (let [first-different-day (transit-changes/first-different-day starting-week-hash
                                                                  different-week-hash)
         starting-week-date (.plusDays (:beginning-of-week starting-week) first-different-day)
-        different-week-date (.plusDays (:beginning-of-week different-week) first-different-day)]
+        different-week-date (.plusDays (:beginning-of-week different-week) first-different-day)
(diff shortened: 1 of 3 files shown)


ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -602,11 +602,11 @@
   (let [first-different-day (transit-changes/first-different-day starting-week-hash
                                                                  different-week-hash)
         starting-week-date (.plusDays (:beginning-of-week starting-week) first-different-day)
-        different-week-date (.plusDays (:beginning-of-week different-week) first-different-day)]
+        different-week-date (.plusDays (:beginning-of-week different-week) first-different-day)
```
rckopteri: nice!

ote/src/cljs/ote/views/admin/interfaces.cljs
```diff
@@ -119,7 +119,8 @@
 
 (defn interface-list [e! app]
   (let [{:keys [loading? results filters]}
-        (get-in app [:admin :interface-list])]
+        (get-in app [:admin :interface-list])
+        fd (js/Date. 0)]
```
rckopteri: what is fd?

jkorpelin: it's meant to be short for "first date", probably not a good symbol name though.

PR #974, opened 2019-04-15
165239255 service form texts

jkorpelin:
# Added
* Ability to pass should-update-check function to form/info

# Changed
* external information info texts based on state.


ote/src/cljs/ote/views/transport_service_common.cljs
```diff
@@ -122,12 +123,14 @@
...
-                  {:target "_blank"}]
-                 (when (flags/enabled? :sea-routes)
-                  [:span
-                   (str (tr [:form-help :external-interfaces-intro-2]) " ")
+                 (when (:road transport-type)
+                   [:span (str (tr [:form-help :external-interfaces-intro-rae-text]) " ")
+                    [linkify "https://www.traficom.fi/fi/asioi-kanssamme/saannollisen-henkiloliikenteen-reitti-ja-aikataulutiedon-digitoiminen"
+                     (str (tr [:form-help :RAE-link-text]) ". ")
+                     {:target "_blank"}]])
+                 (when (and (flags/enabled? :sea-routes) (:sea transport-type))
```
rckopteri: Maybe we could lose this flag entirely or only by removing it from codes (but not from configs)

PR #981, opened 2019-04-17
Video false for cypress settings to stop warnings.

jkorpelin:
# Fixed
* some warnings in cypress tests. And possibly remove restrictions from the amount of tests that can be ran


aws/ansible/jenkins/jobs/cypress.groovy
```diff
@@ -27,6 +27,6 @@ job('Cypress-e2e-tests') {
         shell('CYPRESS_RECORD_KEY=${vault_cypress_record_key} '+
               'CYPRESS_NAP_LOGIN=${vault_cypress_nap_username} '+
               'CYPRESS_NAP_PASSWORD=${vault_cypress_nap_password} '+
-              '$(npm bin)/cypress run --browser chrome --record')
+              '$(npm bin)/cypress run --browser chrome --record false')
```
philosotractor: Hm why does this pop up in the diff even though change was made also in 
https://github.com/finnishtransportagency/mmtis-national-access-point/pull/979/files ...

PR #987, opened 2019-04-23
164597097 use holidays

jkorpelin:
# Changed
* Use the "detection-holidays" table in the change-detection.
* Responses for the admin panels csv importing.

### DIFF
--- a/ote/src/clj/ote/services/admin.clj
+++ b/ote/src/clj/ote/services/admin.clj
@@ -403,16 +403,15 @@
           (when (not-empty response-csv)
             (empty-old-exceptions db)
             (http/transit-response (save-exception-days-to-db db response-csv))))
-        {:status 403
+        {:status 500
          :headers {"Content-Type" "application/json+transit"}
-         :body (clj->transit {:status 500
-                              :error "http://traffic.navici.com/tiedostot/poikkeavat_ajopaivat.csv doesn't respond"})}))
+         :body (clj->transit {:status (:status response)
+                              :response (str response)})}))
     (catch Exception e
-      (println "error")
+      (log/warn "Exception csv error: " e)
       {:status 500
        :headers {"Content-Type" "application/json+transit"}
-       :body (clj->transit {:status 500
-                            :error (str e)})})))
+       :body (clj->transit {:error (str e)})})))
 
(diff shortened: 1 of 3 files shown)


ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -309,8 +313,28 @@
...
+        {}
+        dates))))
+
+#?(:clj
+   (defn is-holiday?
+     "Check if given date is a holiday to be skipped. Returns holiday id if it is or nil otherwise."
+     [db date]
+     (let [{::time/keys [month date year]} (time/date-fields date)
+           exceptions (exception-holidays db)]
+       (if (exceptions [date month year])
```
philosotractor: Two calls to `exceptions`, could be made more efficient binding the value using let or rather, `if-let`:

```
(if-let [res (exceptions [date month year])]
    res
   (static-holidays [date month]))

File: ote/src/clj/ote/services/admin.clj [hunk 2]
```diff
@@ -403,16 +403,15 @@
           (when (not-empty response-csv)
             (empty-old-exceptions db)
             (http/transit-response (save-exception-days-to-db db response-csv))))
-        {:status 403
+        {:status 500
```
philosotractor: 👍

ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -309,8 +313,28 @@
...
+    [db]
+    (let [dates (specql/fetch db
+                              :gtfs/detection-holidays
+                              #{:gtfs/date :gtfs/reason}
+                              {:gtfs/date op/not-null?})]
+      (reduce
+        (fn [map holiday]
+          (let [{::time/keys [month date year]} (time/date-fields (:gtfs/date holiday))]
+            (assoc map
+              [date month year] (keyword (str/lower-case (:gtfs/reason holiday))))))
```
philosotractor: Was it handled somewhere if `holiday` is null, e.g. `""` in CSV? I.e. can this rely that holiday name string is guaranteed to be valid.

jkorpelin: yes by default it's just a string of "holiday" if the third arg from the csv is empty.

PR #989, opened 2019-04-23
Upload failing cypress screenshots to s3

erno:


erno:
A remaining buglet is that the echo'd url is not correctly escaped, and may fail when copypasted to browser.

PR #994, opened 2019-04-25
165380488 reports v1

rckopteri:
# Added
* Drop down list to select tertiili or month reports 
   
# Fixed
* Service search total company count
* Admin reports colors
* Admin reports help texts


ote/resources/public/language/en.edn
```diff
@@ -920,7 +920,8 @@ You can also draw the operating area or point to the map with drawing tools. You
  :admin-page
  {:business-id-filter {:operators "Transport operators" :services "Services" :ALL "All"}
   :published-types {:YES "Published" :NO "Not published" :ALL "All"}
-  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "All"}}
+  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "All"}
+  :report-types {:tertiili "Tertiili" :month "Kuukausi"}}
```
Trpal: Should `:tertiili` be in english? `:tertile`

ote/resources/public/language/sv.edn
```diff
@@ -924,7 +924,8 @@ Tjänsterna beskrivs var för sig på egen blankett. I exemplet fyller man i bå
  :admin-page
  {:business-id-filter {:operators "Palveluntuottajista" :services "Palveluista" :ALL "Kaikki"}
   :published-types {:YES "Julkaistu" :NO "Julkaisematta" :ALL "Kaikki"}
-  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "Kaikki"}}
+  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "Kaikki"}
+  :report-types {:tertiili "Tertiili" :month "Kuukausi"}}
```
jkorpelin: Could `:tertiili` be in english? `:tertile`

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -1,9 +1,10 @@
-(ns ote.views.monitor
+(ns ote.views.admin.monitor
```
philosotractor: Should also controller be renamed to match view, i.e. `ote.app.controller.monitor` to `ote.app.controller.admin.monitor` ?

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -1,9 +1,10 @@
...
   (:require [reagent.core :as r]
             [ote.ui.common :as ui-common]
             [ote.localization :refer [tr tr-key]]
-            [ote.app.controller.monitor :as controller]
-            [cljsjs.chartjs]))
+            [ote.app.controller.monitor :as monitor-controller]
```
philosotractor: While this is a matter of convention and taste, 
I'd prefer shorter `:as` aliases to save a bit of space in the source code file. 
We don't have an agreed convention, but a common one is to use prefix/suffixes like 
controller=>ctrl, view=>vw.

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -64,10 +65,10 @@
                                               (update comp)))
                      :component-did-update update})))
 
-(defn monitor-main [e! {:keys [page monitor-data] :as app}]
-  (e! (controller/->QueryMonitorReport))
-  (if monitor-data
-    (let [translate-typekw (fn [type]
+(defn monitor-main [e! app]
+  (if (get-in app [:admin :monitor :monitor-data])
+    (let [monitor-data (get-in app [:admin :monitor :monitor-data])
```
philosotractor: Two consecutive calls to same function are redundant, should be replaced using either `if-let`or by moving the let of `monitor-data` symbol outside the if so it can refer to `monitor-data` instead of making the same function call.

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -81,55 +82,68 @@
...
                                        :datasets [{:data (mapv :count (:companies-by-service-type monitor-data))
                                                    :backgroundColor ["rgb(0,170,187)"
-                                                                     "rgb(102,214,184)"
                                                                      "rgb(102,204,102)"
+                                                                     "rgb(153,204,0)"
                                                                      "rgb(221,204,0)"
                                                                      "rgb(255,136,0)"
                                                                      "rgb(255,102,153)"
-                                                                     "rgb(235,102,204)"]
+                                                                     "rgb(153,0,221)"]
```
philosotractor: Shouldn't all colors be defined in style file, colors.cljs

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -81,55 +82,68 @@
...
           monthly-types (:monthly-types monitor-data)
-          tertiili-types (:tertiili-types monitor-data)]
-      (fn []
-        [:div
-         [:h1 " Valvontanäkymä "]
-         [:p "Valvontanäkymässä voit tarkastella liikkumispalveluiden tarjoajien kokonaismäärän
+          tertiili-types (:tertiili-types monitor-data)
+          chart-type (get-in app [:admin :monitor :report-type])]
+      [:div
+       [:h1 " Valvontanäkymä "]
```
philosotractor: Why spaces around title?

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -81,55 +82,68 @@
...
+          :options [:tertiili :month]}
+         (get-in app [:admin :monitor :report-type])]]
 
-         [:h2 " Liikkumispalveluiden tarjoajien lukumäärä jaoteltuna liikkumispalvelutyypin mukaan kuukausittain"]
-         [:p "Yhdellä liikkumispalveluita tarjovalla palveluntuottajalla voi olla monta erilaista palvelua. Tästä syystä
-         palvelutyypeittäin jaottelu tuottaa eri lukumäärän palveluntuottajia."]
-         [ui-common/linkify " /admin/reports/monitor/csv/monthly-companies-by-service-type " " Lataa kuvaajan tiedot CSV:nä "]
-         [:div {:style {:width "75%"}}
-          [barchart-inner " bar-type-by-month " "right" monthly-types]]
+       (when (= chart-type :month)
```
philosotractor: When only one case should be executed, it's it's more logical and also better performance to use `case` instead of multiple independent condition structures.
So instead of 
`when 1 a
when 2 b
when 3 c`
==>
`case
         1 a
         2 b
         3 c
         :else`

database/src/main/resources/db/migration/V1_147__all_companies_again.sql
```diff
@@ -0,0 +1,92 @@
...
+         g.created as created,
+         'associated_service_operator' as "source",
+         ts."sub-type" as "sub-type"
+    FROM
+         "associated-service-operators" a,
+         "transport-operator" top,
+         "group" g,
+         "transport-service" ts
+   WHERE
+         g.id = top."ckan-group-id"
```
philosotractor: Just confirming: 
this has a condition for group.id = transport-operator.ckan-group-id,
so this doesn't need a filtering condition for all the cross-joined tables like also for transport-service?

ote/src/clj/ote/services/service_search.sql
```diff
@@ -34,20 +34,9 @@ SELECT COUNT(id) FROM "transport-service" WHERE published IS NOT NULL;
...
-         WHERE ts.published IS NOT NULL
-           AND op."deleted?" = FALSE
-        UNION
-        SELECT DISTINCT (x.c)."business-id"
-          FROM (SELECT unnest(COALESCE(sc.companies, ts.companies)) c
-                  FROM "transport-service" ts
-                  LEFT JOIN service_company sc ON sc."transport-service-id" = ts.id
-                 WHERE ts.published IS NOT NULL) x) y;
+-- Count the total number of companies providing transport services or participating in providing
+SELECT COUNT(x."business-id")
```
philosotractor: 👍 

Just mentioning that this seems to be faster (~75ms) to execute than the often referred alternative
```
    SELECT COUNT (DISTINCT "business-id")
    FROM "all-companies";
```
which executes around 88-100ms

ote/src/clj/ote/services/service_search.sql
```diff
@@ -34,20 +34,9 @@ SELECT COUNT(id) FROM "transport-service" WHERE published IS NOT NULL;
...
-          JOIN "transport-operator" op ON ts."transport-operator-id" = op.id
-         WHERE ts.published IS NOT NULL
-           AND op."deleted?" = FALSE
-        UNION
-        SELECT DISTINCT (x.c)."business-id"
-          FROM (SELECT unnest(COALESCE(sc.companies, ts.companies)) c
-                  FROM "transport-service" ts
-                  LEFT JOIN service_company sc ON sc."transport-service-id" = ts.id
-                 WHERE ts.published IS NOT NULL) x) y;
+-- Count the total number of companies providing transport services or participating in providing
```
philosotractor: 👍 

Just mentioning that this seems to be faster (~75ms) to execute than the often referred alternative
```
    SELECT COUNT (DISTINCT "business-id")
    FROM "all-companies";
```
which executes around 88-100ms

ote/resources/public/language/sv.edn
```diff
@@ -924,7 +924,8 @@ Tjänsterna beskrivs var för sig på egen blankett. I exemplet fyller man i bå
  :admin-page
  {:business-id-filter {:operators "Palveluntuottajista" :services "Palveluista" :ALL "Kaikki"}
   :published-types {:YES "Julkaistu" :NO "Julkaisematta" :ALL "Kaikki"}
-  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "Kaikki"}}
+  :interface-formats {:GTFS "GTFS" :Kalkati.net "Kalkati.net" :ALL "Kaikki"}
```
philosotractor: The uppercase keyword names used here even before the commit are a bit out of convention and also not encouraged by https://github.com/bbatsov/clojure-style-guide#naming

ote/resources/public/language/sv.edn
rckopteri: Changed

ote/src/cljs/ote/views/admin/monitor.cljs
rckopteri: Maybe, but we don't have own .admin. namespace for admin stuff yet. So maybe I do this in some other pr.

ote/src/cljs/ote/views/admin/monitor.cljs
rckopteri: I'm not happy with this name also. But I don't like short names that don't tell what they are. ctrl-monitor would be better name, but then we need to change all the names we have been using. :)

database/src/main/resources/db/migration/V1_147__all_companies_again.sql
```diff
@@ -0,0 +1,92 @@
+-- Use UNION to take only unique companies (name,id,created,source) into account.
```
erno: This could use a comment about what the view is for and what it does. Something along the lines of, (i don't know if this is correct) "exposes company information from 3 different sources: 1) ckan group tables joined to transport-operator table 2) associated-services table (read from csv and not registered in NAP service), 3) transport-operator defined companies not registered in ckan"

ote/src/cljs/ote/views/admin/monitor.cljs
rckopteri: changed

ote/src/cljs/ote/views/admin/monitor.cljs
rckopteri: It will make code much readable. I'll make the change.

ote/src/cljs/ote/views/admin/monitor.cljs
rckopteri: Changed to use one if

database/src/main/resources/db/migration/V1_147__all_companies_again.sql
rckopteri: - group table is joinded using transport-operator table
- transport-service table is joinded using associated-service-operators table
- transport-operator table is joinded using associated-service-operators table

ote/src/clj/ote/services/service_search.sql
rckopteri: I can't replicate your results. In Test Server both gives same times about 50ms.
But in production this:
    SELECT COUNT (DISTINCT "business-id")
    FROM "all-companies";

is about 20ms slower.

database/src/main/resources/db/migration/V1_147__all_companies_again.sql
rckopteri: Comment updated

ote/src/clj/ote/services/service_search.sql
philosotractor: FYI Results in the first comment got using _VPN_+Production

philosotractor: updated the comment, meant to say that commit's approach is the faster one.

ote/src/cljs/ote/views/admin/monitor.cljs
philosotractor: Yes. "Find & Replace > replace all" hides the pain ;)

PR #995, opened 2019-04-25
165575508 transit change detection to ignore holidays between no traffic

philosotractor:
# Fixed
* transit-changes: on no-traffic detection ignore holiday keyworded days
* transit-changes: new test for route start
* transit-changes: cleanup old test case code
* transit-changes: detection code formatting

### DIFF
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -657,7 +657,7 @@
                     :changes (compare-route-days-all-changes-for-week db service-id route-key detection-result))
                   detection-result))
               routes)
-        res (into [] (expand-day-changes route-day-changes))]
+        res (vec (expand-day-changes route-day-changes))]
     res))
 
 (defn- date-in-the-past? [^LocalDate date]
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -196,7 +196,7 @@
   (->> (if beginning?
          weekhash
          (reverse weekhash))
-       (take-while nil?)
+       (take-while #(or (nil? %) (keyword? %)))
```
rckopteri: This might be too simple solution, but lets have a look how well it works. No need to make this complicated if simpler is good enough.

philosotractor: What this does, is it makes the logic to treat holidays like no-traffic days.
The current problems are solved, because logic does not think anymore that holidays before route start or after route end, are traffic.
Currently logic already handles time before/after traffic as no-traffic, so holidays will blend in there.

The downside in this might be that a no-traffic+holidays will in some cases reach the no-traffic reporting threshold, when the no-traffic alone would not have done that.

The positive side is, a no-traffic with holiday in between is handled as one big no-traffic period, instead of two short ones.

philosotractor: What this does, is it makes the logic to treat holidays like no-traffic days.
The current problems are solved, because logic does not think anymore that holidays before route start or after route end, are traffic.
Currently logic already handles time before/after traffic as no-traffic, so holidays will blend in there.

The downside in this might be that a no-traffic+holidays will in some cases reach the no-traffic reporting threshold, when the no-traffic alone would not have done that.

The positive side is, a no-traffic with holiday in between is handled as one big no-traffic period, instead of two short ones.

PR #996, opened 2019-04-26
163479020 operator e2e tests

rckopteri:
# Added
* New e2e test to test creating new transport operator
   
# Fixed
* Select new transport operator as acting operator by default


cypress/integration/operator_spec.js
```diff
@@ -31,7 +36,68 @@ describe('Operator creation basic tests', function () {
...
+        cy.get('#btn-add-new-transport-operator').click();
+
+        // Business id is already added
+        cy.get('#input-business-id').type('1234567-8');
+        // Check warning
+        cy.contains(testOp1.businessIdExistsWarning);
+
+        // Try creating random business-id
+        cy.get('#input-business-id').clear(); // Clear business-id
+        cy.get('#input-business-id').type(randomBusinessid());
```
philosotractor: While which approach to take is a matter of debate, using a random id may lead to test sometimes failing and if it's rerun, it passes. E.g. if random id outputs '1234567-8'. 
Due to human laziness the issue might be overlooked then.
I'd prefer the apprach that a running test should have always the same result, so the valid business id should be hard-coded.
That way it's also possible to cover systematically the different types of id validation, e.g. length/characters/exists/other.

cypress/integration/operator_spec.js
```diff
@@ -31,7 +36,68 @@ describe('Operator creation basic tests', function () {
...
+        cy.wait(500);
+        cy.contains(newOperatorName);
+
+        // Ensure that other pages have it selected by default
+        cy.visit('/#/pre-notices');
+        cy.contains(newOperatorName);
+        cy.visit('/#/routes');
+        cy.contains(newOperatorName);
+
+        // Delete operator
```
philosotractor: This should ensure also that deletion works and doesn't produce an error.

cypress/integration/operator_spec.js
```diff
@@ -31,7 +36,68 @@ describe('Operator creation basic tests', function () {
...
+
+        // Add name
+        const newOperatorName = randomName('ctest-op-');
+        cy.get('#input-operator-name').type(newOperatorName);
+
+        // Save
+        cy.get('#btn-operator-save').click();
+
+        // Ensure that operator is correctly added - should be selected by default
+        cy.wait(500);
```
erno: Did this break without the wait? Cypress should retry and wait automatically for the .contains to succeed. "Cypress only retries commands that query the DOM: cy.get(), .find(), .contains(), etc" according to the docs.

rckopteri: For some reason I could not get cypress to wait for the server response. I did try without the wait(500). But this was the last way to make it work.

cypress/integration/operator_spec.js
rckopteri: Added

PR #997, opened 2019-04-30
165675643 multiple interfaces db query problem

rckopteri:
# Fixed
* transit-visualization: calendar - when different-week-date is null
* service-route-hashes-for-date-range query when service has more than one interface

### DIFF
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -288,7 +288,7 @@
 
     ;; Run ended without reaching threshold, remove start date
     (nil? no-traffic-run)
-    (dissoc state :no-traffic-start)
+    (dissoc state :no-traffic-start-date)
 
     :default
     state))
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -19,6 +19,8 @@
   "The amount of days after a no-traffic run is detected as a change."
   16)
 
+(def detection-interval-days 14)
+
 (defqueries "ote/transit_changes/detection.sql")
 (defqueries "ote/services/transit_changes.sql")
 
@@ -843,8 +845,8 @@
                                                      (or (nil? change-date)
                                                          (date-in-the-past? (.toLocalDate change-date))))
                                                    (sort-by :gtfs/change-date route-change-infos)))
-          ;; Set change date to future (every 4 weeks at monday) - This is the day when changes are detected for next time
-          new-change-date (time/sql-date (time/native->date (.plusDays (time/beginning-of-week (.toLocalDate (time/now))) 28)))
+          ;; Set change date to future (every 2 weeks at monday) - This is the day when changes are detected for next time
+          new-change-date (time/sql-date (time/native->date (.plusDays (time/beginning-of-week (.toLocalDate (time/now))) detection-interval-days)))
           transit-chg-res (specql/upsert! db :gtfs/transit-changes
                                           #{:gtfs/transport-service-id :gtfs/date}
                                           {:gtfs/transport-service-id service-id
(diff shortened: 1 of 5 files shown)


ote/src/cljs/ote/views/transit_visualization/calendar.cljs
```diff
@@ -116,16 +116,17 @@
...
-                     [:span (stylefy/use-style {;; nowrap for the "3 pv" part to prevent breaking "pv" alone to new row.
-                                                :white-space "nowrap"})
-                      (str (time/days-until different-week-date) " " (tr [:common-texts :time-days-abbr]) " ")]
-                     [:span (stylefy/use-style {:color "gray"
-                                                :overflow-wrap "break-word"})
-                      (str "("
-                           (day-of-week-number->text (t/day-of-week (time/js-date->goog-date different-week-date)))
-                           " "
-                           (time/format-timestamp->date-for-ui different-week-date) ")")]])}
+                    (let [ddate (if different-week-date different-week-date (time/now))]
```
philosotractor:
```
(if different-week-date different-week-date (time/now))
```
==>
```
(or different-week-date (time/now))
```

ote/src/cljs/ote/views/transit_visualization/calendar.cljs
```diff
@@ -116,16 +116,17 @@
...
+                      [:div
```
philosotractor: I guess this will do, but commenting that this styling seems similar to what is done in `ote.style.transit_changes.cljs` and referred if transit_changes.cljs and transit_visualization.cljs.
E.g. `transit-changes-icon-row-container`. 
Calendar row style could be defined there as well, if you have the energy.

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -535,4 +534,4 @@
   ;; Timeout used because toggling key for route-changes table directly may cause delay in rendering the content with large data set.
   ;; Thus disabling of UI components must happen before table model change because otherwise table rendering delays those as well.
   (.setTimeout js/window #(e! (->InitiateRouteModelUpdate)) 0)
-  (update app :all-route-changes-checkbox not))
+  app)
```
philosotractor: In this form `ToggleShowNoChangeRoutes` does not contain any functionality, which makes it is useless and should be removed.
As the comments state, event was implemented to separate rendering the checkbox and rendering the state model changes into different js event loop events, because latter one blocks UI in certain cases with large data for some time.

Whichever approach is taken, please check that there's no bad UX using different browsers.

rckopteri: So I did break up the entire fix that you made earlier. I did test this only with services that had small amount of routes. I'll try change it back. Thanx

ote/src/cljs/ote/views/transit_visualization/calendar.cljs
rckopteri: changed

ote/src/cljs/ote/views/transit_visualization/calendar.cljs
rckopteri: This is copy paste from transit_changes route list with minor changes. I do not want to refactor this now. Maybe when we add the change status (or and) comments.

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -843,8 +845,8 @@
                                                      (or (nil? change-date)
                                                          (date-in-the-past? (.toLocalDate change-date))))
                                                    (sort-by :gtfs/change-date route-change-infos)))
-          ;; Set change date to future (every 4 weeks at monday) - This is the day when changes are detected for next time
-          new-change-date (time/sql-date (time/native->date (.plusDays (time/beginning-of-week (.toLocalDate (time/now))) 28)))
+          ;; Set change date to future (every 2 weeks at monday) - This is the day when changes are detected for next time
```
philosotractor: Minor style issue: IMHO constant should define the interval and comments should not explicitly refer to the interval length (2 weeks) anymore. Reason: later someone changes the constant value and is only likely not to update the comments where the symbol was referred to ==> comments will be out of date :)

PR #1000, opened 2019-05-03
165561387 company csv

rckopteri:
# Added
* New button/link component designed for bigger icons and label text
* monitor view: Added button to download csv with all company names and business id's


ote/src/cljs/ote/ui/buttons.cljs
```diff
@@ -51,3 +51,17 @@
...
              opts)
    label])
+
+(defn big-icon-button-with-label
+  "Give icon-element in format: (ic/action-description {:style {:width 30
+                                                        :height 30
+                                                        :margin-right \"0.5rem\"
+                                                        :color colors/primary}})
+  And label-text as text.
+  element-tag is :buttor or :a depending on do you want to move the user to different page or do magic right where she/he is."
```
philosotractor: buttor => button

rckopteri: fixed

ote/src/cljs/ote/views/admin/monitor.cljs
```diff
@@ -101,14 +106,26 @@
...
+       [btn/big-icon-button-with-label
+        {:id "btn-all-companies-csv"
+         :href "/admin/reports/monitor/csv/all-companies"
+         :style {:padding "1rem"}}
+        [ic/action-description {:style {:width 30
+                                          :height 30
+                                          :margin-right "0.5rem"
+                                          :color colors/primary}}]
+        "Lataa liikkumispalveluiden tarjoajien tiedot CSV:nä"
+        :a]]
```
philosotractor: Should `linkify` be used for a href's?

ote/src/cljs/ote/ui/buttons.cljs
```diff
@@ -51,3 +51,17 @@
              (stylefy/use-style style-buttons/svg-button)
              opts)
    label])
+
+(defn big-icon-button-with-label
```
philosotractor: Just wondering, is there a need for a common component to "encapsulate" link or button inside same component?
Both have typically different handling so what's the advantage gained?

ote/src/cljs/ote/ui/buttons.cljs
```diff
@@ -51,3 +51,17 @@
...
+                                                        :margin-right \"0.5rem\"
+                                                        :color colors/primary}})
+  And label-text as text.
+  element-tag is :button or :a depending on do you want to move the user to different page or do magic right where she/he is."
+  [options icon-element label-text element-tag]
+  [element-tag (merge options
+                  (stylefy/use-style style-base/blue-link-with-icon)
+                  (stylefy/use-style style-buttons/outline-button))
+   icon-element
+   [:span {:style {:padding-top "7px"}} label-text]])
```
philosotractor: Isn't the sort-of policy nowadays to have relative units for styles unless there's an explicit reason to have px?

ote/src/cljs/ote/ui/buttons.cljs
rckopteri: Typically buttons are used when something needs to be happening on the same page. And links are used when user is moved to some other location. But in this case csv is downloaded from server and easiest way to do this is via link. But that link is rendered like a button.
Same thing is in own-services page, but without big icon.
Not sure if similar things are in other pages too.

ote/src/cljs/ote/ui/buttons.cljs
rckopteri: Yes. but how do you center texts when they are depending on icon sizes? This way I could make this work.

philosotractor: If both icon and text are ok to be on same x-axis, then parent: `display: flex`, `align-items: center`
refer to e.g. warning-msg which uses `msg-container`

ote/src/cljs/ote/ui/buttons.cljs
philosotractor: Indeed those are the use-cases for link and button elements, but
the question I wasn't sure about was if there's a need for component to encapsulate them, compared to one component which is a link but looks like a button and another component which is a button.
Probably I'm missing something.

rckopteri: Component changed to button

ote/src/cljs/ote/ui/buttons.cljs
rckopteri: fixed

PR #1004, opened 2019-05-06
165815637 constants into one file

philosotractor:
# Changed
* config: transit-changes move route end detection threshold to config.edn


ote/src/clj/ote/tasks/gtfs.clj
```diff
@@ -66,9 +66,9 @@
 ;; To speed up detection, call this with vector of service-ids:
 ;; e.g.: (detect-new-changes-task db (time/now) true [1289])
```
rckopteri: Could you update this example how this can be called from repl? Something like (def config (:config (ote.main/config))

philosotractor: Thanks, done

philosotractor:
Closing because slurping from .edn is indeed a bit overkill for static constants. Will create an another PR with alternative approach.

PR #1018, opened 2019-05-14
165672337 latest changes indicator

jkorpelin:
# Added
* Visualization for recent changes in changes table.

### DIFF
--- a/ote/src/cljs/ote/app/controller/transit_visualization.cljs
+++ b/ote/src/cljs/ote/app/controller/transit_visualization.cljs
@@ -102,11 +102,20 @@
 
         ;; Group by only-changes by route-hash-id
         grouped-changes (group-by :route-hash-id only-changes)
+        group-recent? (map
+                         (fn [[_ changes]]
+                           (some? (some #(:recent-change? %) changes)))
+                         grouped-changes)
         ;; Take first from every vector
         route-changes (map #(first (second %)) grouped-changes)
+        route-changes-with-recent (map
+                                    (fn [change recent?]
+                                      (assoc change :recent-change? recent?))
+                                    route-changes
+                                    group-recent?)
         route-changes (map (fn [x]
                              (assoc x :count (count-changes (:route-hash-id x) grouped-changes)))
-                           route-changes)
+                           route-changes-with-recent)
         sorted-changes (sort-by (juxt :different-week-date :route-long-name :route-short-name) route-changes)
         all-sorted-changes (if show-no-change
                              (concat sorted-changes no-changes)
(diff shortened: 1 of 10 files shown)


ote/src/cljs/ote/views/transit_visualization.cljs
```diff
@@ -353,15 +353,24 @@
...
+       {:name "Reitti" :width "20%"
         :read (juxt :route-short-name :route-long-name)
         :col-style style-base/table-col-style-wrap
         :format (fn [[short long]]
                   (str short " " long))}
 
        ;; Show Reitti/Määränpää column only if it does affect on routes.
        (when (service-is-using-headsign route-hash-id-type)
-         {:name "Reitti/määränpää" :width "23%"
+         {:name "Reitti/määränpää" :width "21%"
```
philosotractor: Just FYI, in this `service-is-using-headsign` case the column gets cramped but I guess this is the only one we can steal some screen real-estate.

ote/src/cljs/ote/style/base.cljs
```diff
@@ -240,6 +240,26 @@
    :padding-right "0"
    :overflow-wrap "break-word"})
 
+(def new-change-indicator
```
philosotractor: `style/transit_changes.cls` is the shared style for `transit-changes` and `transit-visualization`. Should these be moved there?

ote/src/cljs/ote/style/base.cljs
```diff
@@ -240,6 +240,26 @@
...
+  {:height "6px"
```
philosotractor: Should unit format be relative to font height like `rem`?
In extreme case font could be big and icon small.

ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -102,11 +102,20 @@
 
         ;; Group by only-changes by route-hash-id
         grouped-changes (group-by :route-hash-id only-changes)
+        group-recent? (map
+                         (fn [[_ changes]]
+                           (some? (some #(:recent-change? %) changes)))
+                         grouped-changes)
         ;; Take first from every vector
         route-changes (map #(first (second %)) grouped-changes)
+        route-changes-with-recent (map
```
philosotractor: Alternative approach would be to replace `route-changes-with-recent` with by adding into `group-recent?` following assoc
```
 route-changes (map 
  #(let [res (first (second %))]
      (assoc res :recent-change? val) ;; val = true if filter for second returns any items with  :recent-change?
  grouped-changes)
```

PR #1021, opened 2019-05-16
own-service: fixed bug with service errors.

jkorpelin:
Duplicated the service if it had many interfaces.

   
# Fixed
* own services page showed error if a service had multiple interfaces and one of the wasn't type `route-and-schedule`


ote/src/clj/ote/services/transport.sql
```diff
@@ -4,15 +4,18 @@ SELECT description FROM "group" WHERE id = :id;
...
+       ts."name",
+       ts."type",
+       ts."sub-type",
+       ts."published",
+       ts."created",
+       ts."modified",
+       ts."transport-type",
+       (SELECT array_agg(dc)
+        FROM "external-interface-description" eid
+                 JOIN LATERAL unnest(eid."data-content") dc ON TRUE
```
rckopteri: Indentation is sub optimal

PR #1022, opened 2019-05-17
166029512 notification email change count errors

rckopteri:
# Fixed
* Change email: Fixed changed route count
   
# Changed
* transit-visualization: Added route end icon to route list when no-traffic route has ending change also.

### DIFF
--- a/ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
+++ b/ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
@@ -9,7 +9,8 @@
             [ote.ui.icons :as ote-icons]
             [clojure.string :as str]
             [ote.theme.colors :as colors]
-            [ote.localization :refer [tr]]))
+            [ote.localization :refer [tr]]
+            [ote.time :as time]))
 
 ;; Utility methods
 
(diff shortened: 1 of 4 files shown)


ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
```diff
@@ -38,6 +39,17 @@
       changes
       (when with-labels? " aikataulumuutosta")]]))
 
+(defn no-traffic-icons [{:keys [change-type] :as change-summary}]
```
philosotractor: Is `no-traffic-icons` referred to anywhere?

rckopteri: Yes. in transit-visualization.cljs on row 403

philosotractor: Instead of `{:flex "0.5"}` the implicit flex-grow number could be `{:flex "1"}` because it's the ratio that matters.
But this is nitpicking.

jkorpelin: All the flex numbers could be changed to integers with the proper ratios. I don't think floats are all that good for flex.

PR #1023, opened 2019-05-20
165876582 spatial ui subheading

rckopteri:
# Added
* service-search: New subheadings for spatial search

### DIFF
--- a/ote/src/cljs/ote/style/service_search.cljs
+++ b/ote/src/cljs/ote/style/service_search.cljs
@@ -33,7 +33,7 @@
 
 (def result-card {:box-shadow "1px 1px 10px 0 hsla(0, 0%, 86%, .5)"
                   :background-color "#fff"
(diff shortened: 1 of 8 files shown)


ote/src/cljs/ote/style/service_search.cljs
```diff
@@ -107,3 +107,6 @@
 (def contact-icon {:color  "#999999"
                    :height 16
                    :width  16})
+
+(def title-group-description  {:font-size "0.875rem"
+                               :margin-top "-1rem"})
```
jkorpelin: I think it would be better to just give this `:margin-top 0` and remove the bottom margin from the `heading` element above this text.

rckopteri: Heading element is our page/page-controls title. If I change it it will be changed in everywhere. But in other places it should not be done.
So I think that I leave this as it is.

PR #1032, opened 2019-05-24
165675611 admin service endpoint refactoring 2

philosotractor:
# Added
Utility funcs used in REST endpoint implementation:
* http: generic func for returning http error if map spec validation fails
* util: spec validation function with error string output on fail

# Changed
admin REST endpoint refactoring: 
* controller: admin refactoring rename misleading symbol name
* controller: admin user-operator-members api renamed as member, refactoring
* services: admin user-operator-members api renamed as member, refactoring
* views: admin view typo fix

Other minor cleanup
* util: deleted clojure.clj, content moved to ote.util.collections
* services: transit-visualization remove unused ote.util.clojure require
* util: moved remove-coll-key-ns to existing more suitable namespace

### DIFF
--- a/ote/src/clj/ote/services/admin.clj
+++ b/ote/src/clj/ote/services/admin.clj
@@ -34,7 +34,9 @@
             [ote.services.external :as external]
             [ote.tasks.pre-notices :as pn]
             [hiccup.core :refer [html]]
-            [ote.email :as email]))
+            [ote.email :as email]
+            [clojure.spec.alpha :as spec]
+            [ote.util.collections :as ote-coll]))
 
 (defqueries "ote/services/admin.sql")
 (defqueries "ote/services/reports.sql")
@@ -62,7 +64,7 @@
 
 (defn- authorization-fail-response [user]
   (when-not (:admin? user)
-    (log/warn "authorization-fail-response: id=" (:id user))
+    (log/info  "authorization-fail-response: id=" (:id user))
     (http/transit-response "Not authorized. Bad role." 403)))
 
 (defn- admin-service [route {user :user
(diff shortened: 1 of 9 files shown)


ote/src/clj/ote/services/admin.clj
```diff
@@ -469,17 +489,17 @@
 (defn- admin-routes [db http nap-config email-config]
   (routes
 
-    (GET "/admin/user" [search type :as req]
+    (GET "/admin/user" req
       (or (authorization-fail-response (get-in req [:user :user]))
-          (list-users-bad-req-response search type)
-          (list-users-response db search)))
+          (list-users-response db (ote-coll/map->keyed (:params req)))))
```
rckopteri: 👍

ote/src/cljs/ote/app/controller/admin.cljs
```diff
@@ -147,18 +146,19 @@
...
     (assoc-in app [:admin :user-listing :loading?] true))
 
   ConfirmDeleteUser
   (process-event [{id :id} app]
     (if (= id (:ensured-id (get-user-by-id app id)))
       (comm/delete! (str "admin/user/" id)
                     nil
-                    {:on-success (tuck/send-async! ->ConfirmDeleteUserResponse)
-                     :on-failure (tuck/send-async! ->ConfirmDeleteUserResponseFailure)})
+                    {:on-success (tuck/send-async! ->ConfirmDeleteUserResponse)})
```
rckopteri: Why not on-failure case?

philosotractor: During development I added extra on-failures and this was removed by accident while removing those because it seemed to need proper handling. But it's better to keep the old functionality, no time to refactor everything :)

philosotractor: So on-failure restored, thanks!

PR #1033, opened 2019-05-27
166029512 wrong change count

rckopteri:
# Added
* Transit-visualization: Route list - Added two new icons: added route, removed route
   
# Fixed
* Pre-notice email: Fixed different-week-date in "time for the first change" column
* Transit-changes, Transit-visualization: Refactored icon colors
   
# Changed
* Transit-visualization: Route list - added all icons to every route and "disabled" them when route do not have all kinds of changes


ote/src/cljs/ote/app/controller/transit_visualization.cljs
```diff
@@ -114,7 +120,9 @@
                                     route-changes
                                     group-recent?)
         route-changes (map (fn [x]
-                             (assoc x :count (count-changes (:route-hash-id x) grouped-changes)))
+                             (-> x
+                                 (assoc :count (count-changes (:route-hash-id x) grouped-changes))
+                                 (assoc :combined-change-types (combine-change-types (get grouped-changes (:route-hash-id x))))))
```
philosotractor: Does :combined-change-types need to be a string? 
Should we prefer operations using keywords instead of strings?
Would a `set` or `list` suit as well? Then `contains?` etc. queries could use keywords without string operations.
If :change-type is a string in root map, it could be converted here via `:keyword` before conjoining to the collection.

ote/src/cljs/ote/views/transit_visualization/change_icons.cljs
```diff
@@ -39,16 +39,73 @@
...
+     [:div {:style {:flex "1"}}
+      [icon-l/icon-labeled
+       [ic/action-query-builder {:style {:color colors/icon-gray}}]
+       [:span (or trip-stop-time-changes-upper trip-stop-time-changes-lower)]]]
+     [:div {:style {:flex "1"}}
+      [icon-l/icon-labeled
+       [ic/action-query-builder {:style {:color colors/icon-disabled}}]
+       [:span (or trip-stop-time-changes-upper trip-stop-time-changes-lower)]]])
+
+   (if (str/includes? (str combined-change-types) ":no-traffic")
```
philosotractor: Isn't  `combine-change-types` already a string?

PR #1036, opened 2019-05-28
service: admin fix delete user when there are no memberships

philosotractor:
# Fixed
* service: admin fix delete user when there are no memberships


philosotractor:
Some issue in CI/CA, https://github.com/finnishtransportagency/mmtis-national-access-point/pull/1037 replaces this PR

PR #1040, opened 2019-06-04
164101206 add new member

rckopteri:
# Added
* Transport operator members: Add existing user as a member of operator
* Transport operator members: Inform existing user that he/she has been added as a member of operator by sending html email
* Auditlog: Add auditlog when existing user is added as a member of operator
* e2e test: Add test that test if existing user is added as a member of operator


ote/src/cljs/ote/views/operator_users.cljs
```diff
@@ -70,6 +71,6 @@
      [:h2 name]
      (if loaded?
        [:div
-        [access-table e! access-users name]
+        [access-table e! access-users (get-in state [:params :operator-id])]
         [invite-member e! access-state (get-in state [:params :operator-id])]]
-       [spinner/primary-spinner])]))
+       [spinner/primary-spinner :id "waiting-loading-event-spinner"])]))
```
philosotractor: I see a new spinner has been defined (by @jkorpelin) recently. 
However there already was a material-ui spinner `ote.ui.circular_progress` (which is css as well) which is used in a few places. 
There shouldn't be two overlapping implementations.
Shouldn't we use the material-ui component by default in order to avoid re-inventing the wheel?

ote/src/cljs/ote/ui/spinner.cljs
```diff
@@ -73,6 +73,7 @@
           (merge
             (when-let [style (:style options)]
```
philosotractor: Alternative implementation
```suggestion
           (:style options)
```
Because when returns a nil if style isn't defined, so does the suggestion. FYI @jkorpelin

philosotractor: But as commented in an another part: shouldn't the material-ui spinner be used instead of this...

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
...
+(defn remove-member-from-operator [email db user operator-id member-email]
+  (let [member (first (fetch-user-by-email db {:email member-email}))
+        operator (first (specql/fetch db ::t-operator/transport-operator
+                                      #{::t-operator/id ::t-operator/name ::t-operator/ckan-group-id}
+                                      {::t-operator/id operator-id}))]
+    (specql/delete! db ::user/member
+                    {::user/table_id   (:user_id member)
+                     ::user/group_id   (::t-operator/ckan-group-id operator)
+                     ::user/table_name "user"})
+    (http/transit-response "OK")))
```
philosotractor: (In lack of overload which takes only status code)
```suggestion
    (http/transit-response nil 200)))
```
Is a string necessary?

ote/src/clj/ote/services/transport.clj
```diff
@@ -583,7 +663,19 @@
              :params
              user :user
              form-data :body}
-        (http/transit-response (http/transit-request form-data))) ;THIS IS JUST PLACEHOLDER
+            (authorization/with-transport-operator-check
+              db user (Long/parseLong operator-id)
+              #(manage-adding-users-to-operator email db user (Long/parseLong operator-id)
+                                                (http/transit-request form-data))))
+
+      ;; This is not ready, only implemented to help add-member implementation
```
philosotractor: Proposal
```
DELETE /transport-operator/:operator-id/users/?email=myemail@address.com
```
(Although the resource could be `/transport-operator/:operator-id/member/?` but "users" is already used in this domain so...)

FYI route's`params` key contains url query strings as a map ({"email" "myemail@address.com"})
For an example see GET /admin/member?userid

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
...
+    (if (and response (not (:error response)))
+      (http/transit-response response 200)
+      (http/transit-response response 400))))
+
+(defn remove-member-from-operator [email db user operator-id member-email]
+  (let [member (first (fetch-user-by-email db {:email member-email}))
+        operator (first (specql/fetch db ::t-operator/transport-operator
+                                      #{::t-operator/id ::t-operator/name ::t-operator/ckan-group-id}
+                                      {::t-operator/id operator-id}))]
+    (specql/delete! db ::user/member
```
philosotractor: What if `member` or `operator` is not found? Skip executing specql?

rckopteri: Delete members from operator will be implemented in ticket: 165675263. I only made it possible in this pr to enable e2e tests. So no need to review anything related to removing member.

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
...
+                        (invite-new-user email db requester operator (:email form-data))
+
+                        ;; new-member is already a member
+                        (and (not (empty? new-member)) new-member-is-operator-member)
+                        {:error :already-member}
+
+                        ;; Something wrong is happening
+                        :else
+                        {:error :something-went-wrong})]
+    (if (and response (not (:error response)))
```
philosotractor:
```suggestion
    (if (not (:error response))
```

rckopteri: Will change

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
   [db operator-id]
   (http/transit-response (fetch-operator-users db {:operator-id operator-id})))
 
+(defn add-user-to-operator [email db new-member requester operator]
```
philosotractor: Implementation relies that `create-member!` won't fail. 
Is that ok? Should a failing insert throw an exception or return an error response?

ote/src/cljs/ote/ui/spinner.cljs
rckopteri: I change it to: (when (:style options) (:style options))
Because no need to merge when options is null.

ote/src/clj/ote/services/transport.clj
rckopteri: Really good point. I get into that. But it is very unlikely that simple insert will fail.

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
...
+(defn manage-adding-users-to-operator [email db requester operator-id form-data]
+
+  (let [new-member (first (fetch-user-by-email db {:email (:email form-data)}))
+        operator (first (specql/fetch db ::t-operator/transport-operator
+                                      #{::t-operator/id ::t-operator/name ::t-operator/ckan-group-id}
+                                      {::t-operator/id operator-id}))
+        operator-users (fetch-operator-users db {:operator-id operator-id})
+        new-member-is-operator-member  (some #(= (:user_id new-member) (:id %)) operator-users)
+        ;; If member exists, add it to the organization if not, invite
+        response (cond
```
philosotractor: Indentation could be adjusted :)

rckopteri: Thanx, fixed

ote/src/clj/ote/services/transport.clj
```diff
@@ -177,11 +176,11 @@
      :transport-service-vector transport-services-vector
      :user cleaned-user}))
 
-(defn- create-member! [db user-id group]
+(defn- create-member! [db user-id group-id]
   (specql/insert! db ::user/member
                   {::user/id         (str (UUID/randomUUID))
                    ::user/table_id   user-id
-                   ::user/group_id   (:ote.db.transport-operator/group-id group)
+                   ::user/group_id   group-id
```
philosotractor: 👍

ote/src/clj/ote/nap/users.sql
```diff
@@ -13,6 +13,21 @@
  LEFT JOIN "group" g ON g.id = m.group_id
      WHERE u.name = :username;
 
+-- name: fetch-user-by-email
```
philosotractor: There are <number not printed here> of users with `null` email. This may therefore return a set of records, which is not intended.

ote/resources/public/language/fi.edn
```diff
@@ -601,7 +601,9 @@ Voit myös piirtää toimialueen tai toimipisteen kartan piirtotyökalujen avull
                         :members "Jäsenet"
                         :add-member "Lisää jäsen"
                         :email-placeholder "Kirjoita sähköpostiosoite"
-                        :manage-users "Hallinnoi jäseniä"}
+                        :manage-users "Hallinnoi käyttöoikeuksia"
+                        :something-went-wrong "Jotain meni pieleen."
```
philosotractor: Could be a generic error message which could be reused? Not an issue really.

ote/src/clj/ote/nap/users.sql
rckopteri: It is not possible to have null email and state='active'. Or maybe it is theoretically but in production there are non that are valid user.

ote/src/clj/ote/services/transport.clj
```diff
@@ -541,9 +540,90 @@
   [db operator-id]
   (http/transit-response (fetch-operator-users db {:operator-id operator-id})))
 
+(defn add-user-to-operator [email db new-member requester operator]
+  (let [group (create-member! db (:user_id new-member) (::t-operator/ckan-group-id operator))
+        ;; Ensure that we do not send email to test user
+        updated-member (if (= "user.userson@example.com" (:user_email new-member))
```
philosotractor: Hard-coded values should be put to named symbols. Tests and code could then refer to same symbol, allowing updating value to one row.
Those constant symbols should be defined e.g. in cljc/ote.config/something.

Would logic readability be improved by moving the test user detection logic (one row, I admit) to separate function to make it more obvious what's happening?

ote/src/clj/ote/nap/users.sql
philosotractor: Thus select "where" should check that
state="active" AND email IS NOT NULL

ote/src/cljs/ote/views/operator_users.cljs
rckopteri: I made a ticket for this. Lets figure out what spinner we will use in the future.

ote/src/clj/ote/services/transport.clj
rckopteri: Nope. Will be fixed in remove member ticket.

ote/src/clj/ote/services/transport.clj
rckopteri: Same logic is used earlier so I think that this is ok for now.

ote/src/clj/ote/services/transport.clj
rckopteri: Added email-config.cljc file for these hard coded addresses

ote/src/clj/ote/services/transport.clj
philosotractor: Ok. An easy way out would be to throw an exception if nil member is returned ==> http request completed automagically with status 500 or similar.

PR #1043, opened 2019-06-05
166438650 remove expired detected changes from transit-changes list

rckopteri:
# Fixed
* transit-changes page: Recalculate change amounts for all services every night

### DIFF
--- a/ote/src/clj/ote/services/transit_visualization.sql
+++ b/ote/src/clj/ote/services/transit_visualization.sql
@@ -182,7 +182,7 @@ WHERE c."transport-service-id" = :service-id
 -- Fetch changes for service joinin detected-route-change table and detected-change-history table.
 SELECT h."change-detected", c."route-short-name", c."route-long-name", c."trip-headsign", c."route-hash-id", c."change-type", c."added-trips",
        c."removed-trips", c."trip-stop-sequence-changes-lower", c."trip-stop-sequence-changes-upper",
-       c."trip-stop-time-changes-lower", c."trip-stop-sequence-changes-upper", c."current-week-date",
+       c."trip-stop-time-changes-lower", c."trip-stop-time-changes-upper", c."current-week-date",
        c."different-week-date", c."change-date", c."created-date",
        h."email-sent" >= (SELECT MAX("email-sent") FROM "detected-change-history") AS "recent-change?"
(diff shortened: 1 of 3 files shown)


ote/src/clj/ote/services/transit_visualization.sql
```diff
@@ -182,7 +182,7 @@ WHERE c."transport-service-id" = :service-id
 -- Fetch changes for service joinin detected-route-change table and detected-change-history table.
 SELECT h."change-detected", c."route-short-name", c."route-long-name", c."trip-headsign", c."route-hash-id", c."change-type", c."added-trips",
        c."removed-trips", c."trip-stop-sequence-changes-lower", c."trip-stop-sequence-changes-upper",
-       c."trip-stop-time-changes-lower", c."trip-stop-sequence-changes-upper", c."current-week-date",
+       c."trip-stop-time-changes-lower", c."trip-stop-time-changes-upper", c."current-week-date",
```
philosotractor: Good catch

ote/src/clj/ote/tasks/gtfs.clj
```diff
@@ -105,6 +106,34 @@
...
+(defn recalculate-detected-changes-count
+"Change detection (detect-new-changes-task) stores detected changes. These changes will expire after some period of time.
+All changes occur on a certain day and they will get old. Transit-changes page shows how many changes there are and
+if we don't recalculate them every night expired changes will be show also. So in this fn we recalculate change counts again using
+different-week-date value and skip all expired changes."
+  [db]
+  (let [upcoming-changes (upcoming-changes db)
+        _ (log/info "Detected change count recalculation started!")]
+    ;; Loop upcoming-changes
+    (dotimes [i (count upcoming-changes)]
```
philosotractor: Ok like this of courser,
food for thought: would a more functional approach be to loop upcoming-changes and per iteration take `first` and recur `rest`?
loop syntax is always a bit verbose :/

rckopteri: I'm not sure if this is such a bad option because to enable loop with index number there will be extra code.

philosotractor: loop wouldn't use index, it would use only `first` and `rest`.
But anyways...

PR #1046, opened 2019-06-06
Ports csv report

rckopteri:
# Added
* New csv report to admin panel: Ports

### DIFF
--- a/ote/src/clj/ote/services/admin.clj
+++ b/ote/src/clj/ote/services/admin.clj
@@ -499,7 +499,7 @@
       (log/warn "Error while sending a notification" e))))
 
 
-(defn- all-ports [db]
+(defn- all-ports-response [db]
   (csv-data ["Koodi" "Nimi" "Leveyspiiri (lat)" "Pituuspiiri (lon)" "Käyttäjän lisäämä?" "Luontihetki"]
             (map (juxt :code :name :lat :long :user-added? :created)
                  (fetch-all-ports db))))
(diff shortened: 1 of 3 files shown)


ote/src/clj/ote/services/admin.clj
```diff
@@ -577,7 +582,16 @@
        {{:keys [type]} :params
         user :user}
     (require-admin-user "reports/transport-operator" (:user user))
-    (transport-operator-report db type)))
+    (transport-operator-report db type))
+
+
+  ^{:format :csv
+    :filename (str "raportti-" (time/format-date-iso-8601 (time/now)) ".csv")}
+  (GET "/admin/reports/all-ports"
```
philosotractor:
```suggestion
  (GET "/admin/reports/port"
```

ote/src/clj/ote/services/admin.clj
```diff
@@ -577,7 +582,16 @@
...
+       {{:keys [type]} :params
```
philosotractor: This is not used?
```suggestion
       {{:keys [type]} :params
```

ote/src/clj/ote/services/admin.clj
```diff
@@ -577,7 +582,16 @@
...
+        user :user}
+    (require-admin-user "reports" (:user user))
```
philosotractor: Isn't `require-admin-user` nowadays sorta-kinda deprecated for us, should `authorization-fail-response` style design be preferred?
==> rename/refactor also `all-ports` to `all-ports-response` ?

ote/src/clj/ote/services/admin.clj
rckopteri: Renamed

ote/src/clj/ote/services/admin.clj
rckopteri: Removed

ote/src/clj/ote/services/admin.clj
rckopteri: Changed

PR #1048, opened 2019-06-07
166535965 rest end point refactoring

rckopteri:
# Changed
* Refactored some rest end points


ote/src/cljs/ote/app/routes.cljs
```diff
@@ -130,7 +144,9 @@
                      orig-app app
                      app (merge app navigation-data)]
 
-                 (if (requires-authentication? app)
+                 (if (or (requires-authentication? app)
```
philosotractor: Should functions return true if user passes a check and fail if not?
Should functions be renamed based on the check they make?

```suggestion
                 (if (or (not (page-user-authenticated? app))
                         (not (page-user-transit-authority-authorized? app))
                         (not (page-user-admin-authorized app))
```

ote/src/clj/ote/authorization.clj
```diff
@@ -6,6 +6,10 @@
             [taoensso.timbre :as log]
             [ote.nap.users :as users]))
 
+(defn require-transit-authority [user]
```
philosotractor: So there's two options to handle a HTTP request fail: throwing and exception or returning a response object?
I suppose both are ok and it's a matter of design which is used in the app.

Admin refactoring uses the latter approach: a function is called to resolve an authentication fail response map, if it's not returned then a response is requested from a domain function.
Should same design be used here? 
services.admin uses `authorization-fail-response` to return `(http/transit-response "Not authorized. Bad role." 403)` which could be used here as well.
Additional upside there is that request status code is explicitly defiened, with `SecurityException` you have to find out what it will be.

ote/src/cljs/ote/views/main.cljs
```diff
@@ -82,14 +81,14 @@
         [:div {:style (stylefy/use-style style-base/body)}
          [theme e! app
           (if (nil? app)
-            [circular-progress/circular-progress]
+            [spinner/circular-progress]
```
philosotractor: Are these spinner changes included on purpose?

ote/src/cljs/ote/app/routes.cljs
rckopteri: Functions returns boolean value true/false. It is indicated using ? in the function name. I think that functions are named well enough and for this time I'm not changing their names.

ote/src/clj/ote/authorization.clj
rckopteri: My bad. Will change.

ote/src/cljs/ote/views/main.cljs
rckopteri: No. but they won't hurt :)

PR #1052, opened 2019-06-12
166538796 reorder week days in service view

rckopteri:
# Added
* Add function that reorders week day set. And take it into use in service-viewer page

### DIFF
--- a/ote/src/cljc/ote/util/transport_service_util.cljc
+++ b/ote/src/cljc/ote/util/transport_service_util.cljc
@@ -1,0 +1,20 @@
+(ns ote.util.transport-service-util
+  "Util functions for transport service")
+
+(def week-day-order {:MON 1, :TUE 2, :WED 3, :THU 4, :FRI 5, :SAT 6, :SUN 7})
+
+;; Reorder week days mon 1, tue 2, wed 3, thu 4, fri 5, sat 6, sun 7
+(defn reorder-week-days
+  "Receives set of week days. Returns list of week days to make sure that ordering remains correct."
+  [week-day-set]
+  (let [ordered-days (sort (mapv #(% week-day-order) week-day-set))
+        ordered-days->list (map
+                             (fn [od]
+                               (first
+                                 (first
+                                   ;; Return map which value equals ordered week day number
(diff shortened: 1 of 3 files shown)


ote/src/cljc/ote/util/transport_service_util.cljc
```diff
@@ -0,0 +1,20 @@
...
+(def week-day-order {:MON 1, :TUE 2, :WED 3, :THU 4, :FRI 5, :SAT 6, :SUN 7})
+
+;; Reorder week days mon 1, tue 2, wed 3, thu 4, fri 5, sat 6, sun 7
+(defn reorder-week-days
+  "Receives set of week days. Returns list of week days to make sure that ordering remains correct."
+  [week-day-set]
+  (let [ordered-days (sort (mapv #(% week-day-order) week-day-set))
+        ordered-days->list (map
+                             (fn [od]
+                               (first
```
jkorpelin: double `(first)` call could be replaced with `(second)`?

jkorpelin:
I'm not sure if this is needed since the days could be sorted when saved? And the days are sorted in the form where they are added, could the same method be used in the all info page?

rckopteri:
Response for Joni: If we are dependent on the order where week days are saved, then the ordering should be in the back end when they are saved. Also we should reorder all days that are already in the database.
So the simplest way is to reorder them when they are shown.

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -466,7 +467,8 @@
              (tr [:service-viewer :day-of-week])
              (string/join ", " (map
                                  #(string/lower-case (tr [:enums ::t-service/day :short %]))
-                                 (::t-service/week-days time))) true]]
+                                 (tsu/reorder-service-days (::t-service/week-days time))))
```
philosotractor:
```suggestion
                                 (tsu/reorder-week-days (::t-service/week-days time))))
```
?

rckopteri: hups. Thanx!

PR #1055, opened 2019-06-15
166495087 transit changes no-traffic period logic refactoring

philosotractor:
# Changed
* transit-changes: tests, new cases for changes after no-traffic
* transit-changes: detection_test_utilities create-data-all-routes
* transit-changes: change_history, safety check
* transit-changes: detection, generic :holiday keyword for holiday hashes
* transit-changes: detection, refactor changes-by-week->changes-by-route
* transit-changes: detection, refactor no-transit detection
* transit-changes: simplify no-traffic detection logic

Let's plan whether to merge this to master and then https://github.com/finnishtransportagency/mmtis-national-access-point/pull/1008, or if these should be merged together before merging to master.

### DIFF
--- a/ote/src/clj/ote/transit_changes/detection.clj
+++ b/ote/src/clj/ote/transit_changes/detection.clj
@@ -922,25 +922,29 @@
   Function splits the input maps so that routes are not grouped together, instead each route will be in its own collection.
   Output: A vector of 'route-vectors'. Each route-vector contains maps, each representing a week for the specific route.
     Output format:
-       [[{:beginning-of-week 1.1, :end-of-week 7.1, :routes {\"route1\" [\"h1\" \"h1\"]}}
-        {:beginning-of-week 8.1, :end-of-week 15.1, :routes {\"route1\" [\"h1\" \"h1\"]}}
-        {:beginning-of-week 16.1, :end-of-week 23.1, :routes {\"route1\" [\"h1\" \"h1\"]}}]
-        [{:beginning-of-week 1.1, :end-of-week 7.1, :routes {\"route2\" [\"h1\" \"h1\"]}}
-         {:beginning-of-week 8.1, :end-of-week 15.1, :routes {\"route2\" [\"h1\" \"h1\"]}}
-         {:beginning-of-week 16.1, :end-of-week 23.1, :routes {\"route2\" [\"h1\" \"h1\"]}}]]"
-  [weeks]
-  (vals (group-by (fn [d]
-                    (keys (:routes d)))
-                  (reduce
-                    (fn [result week]
-                      (let [routes (:routes week)
-                            r-weeks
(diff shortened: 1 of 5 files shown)


ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -193,14 +189,12 @@
...
-         weekhash
-         (reverse weekhash))
-       (take-while #(or (nil? %) (keyword? %)))
-       count))
+(defn week-hash-key-ix
+  "Input: weekhash = sequence of string hashes
+    key-to-find = key to find
+  Output: Returns the index of first occurrence of `key-to-find`. Monday = 0, Sunday = 6, nil = not found"
+  [weekhash key-to-find]
+  (key-to-find (zipmap weekhash (range 0 8))))
```
rckopteri: (range 8) is the same as (range 0 8)

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -416,21 +341,157 @@
   (assert (some? (:beginning-of-week rw)) rw)
   (not (local-date-before? (:beginning-of-week rw) date)))
 
-(defn filter-by-vals [pred map]
-  (into {} (for [[k v] map
-                 :when (pred v)]
-             [k v])))
+(defn- remove-starting-nt-first
```
rckopteri: replace would be better than remove

ote/src/clj/ote/transit_changes/detection.clj
```diff
@@ -838,6 +881,30 @@
...
+                                (second %))
+                             all-routes)
+            route-min-date-local (when route-info
+                                   (.toLocalDate (:min-date route-info)))
+            route-max-date-local (when route-info
+                                   (.toLocalDate (:max-date route-info)))]
+
+        (if (and route-min-date-local route-max-date-local)
+          (filterv #(and (.isAfter (:end-of-week %) route-min-date-local)
+                         (.isBefore (:beginning-of-week %) route-max-date-local))
```
rckopteri: I'm assuming that localdate don't have time (hours) so when getting isBefore :beginning-of-week might be same day, but with time (hours) and the comparing works even if they are pointing to the same day?

philosotractor: Good catch, will change.

ote/src/clj/ote/transit_changes/detection.clj
philosotractor: True, renamed it a bit later. Will push.

ote/src/clj/ote/transit_changes/detection.clj
philosotractor: True, intentionally used this for clarity, but can remove.

PR #1061, opened 2019-06-20
Fix csv report sql queries

rckopteri:
# Fixed
* Fixed 2 csv reports that were broken because of wrongly made user table join.

### DIFF
--- a/ote/src/clj/ote/services/reports.sql
+++ b/ote/src/clj/ote/services/reports.sql
@@ -3,12 +3,12 @@
 -- name: fetch-operators-no-services
 SELECT op.name, op.id, op.phone, COALESCE(NULLIF(op.email, ''), u.email) AS "email"
   FROM "transport-operator" op
-         JOIN "user" u ON u.name = (SELECT u.name
+         JOIN "user" u ON u.name = (SELECT usr.name
                                       FROM "transport-operator" t
                                              JOIN member m ON m.group_id = t."ckan-group-id"
-                                             JOIN "user" u ON m.table_id = u.id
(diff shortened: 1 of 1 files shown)


ote/src/clj/ote/services/reports.sql
```diff
@@ -3,10 +3,12 @@
 -- name: fetch-operators-no-services
 SELECT op.name, op.id, op.phone, COALESCE(NULLIF(op.email, ''), u.email) AS "email"
   FROM "transport-operator" op
-         JOIN "user" u ON u.name = (SELECT author
-                                      FROM "revision" r
-                                             JOIN "group" g ON op."ckan-group-id" = g.id
-                                     WHERE r.id = g."revision_id"
+         JOIN "user" u ON u.name = (SELECT u.name
```
philosotractor: Ok, but for clarity it would be good to not to use "`u`" twice as alias but for different tables.

ote/src/clj/ote/services/reports.sql
```diff
@@ -133,11 +135,13 @@ SELECT x.*,
                    AND ts.published IS NULL) AS "unpublished-services-count"
 
           FROM "transport-operator" op
-          JOIN "user" u ON u.name = (SELECT author
-                                       FROM "revision" r
-                                              JOIN "group" g ON op."ckan-group-id" = g.id
-                                      WHERE r.id = g."revision_id"
-                                      LIMIT 1)
+          JOIN "user" u ON u.name = (SELECT u.name
```
philosotractor: Ok, but for clarity it would be good to not to use "`u`" twice as alias but for different tables.

PR #1063, opened 2019-06-24
152601167 Add "Back to Draft" button to service form

rckopteri:
# Added
* Add "Back to Draft" button to service form

### DIFF
--- a/ote/src/cljs/ote/views/transport_service_common.cljs
+++ b/ote/src/cljs/ote/views/transport_service_common.cljs
@@ -419,9 +419,15 @@
 
         (if published?
           ;; True
-          [buttons/save {:on-click #(e! (ts/->SaveTransportService schemas true))
-                         :disabled (not (form/can-save? data))}
-           (tr [:buttons :save-updated])]
+          [:span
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/views/transport_service_common.cljs
```diff
@@ -419,9 +419,15 @@
...
         (if published?
           ;; True
-          [buttons/save {:on-click #(e! (ts/->SaveTransportService schemas true))
-                         :disabled (not (form/can-save? data))}
-           (tr [:buttons :save-updated])]
+          [:span
+           [buttons/save {:on-click #(e! (ts/->SaveTransportService schemas true))
+                          :disabled (not (form/can-save? data))}
+            (tr [:buttons :save-updated])]
+           [buttons/save {:disabled name-missing?
```
philosotractor: After pressing :back-to-draft once, save should be enabled and back-to-draft disabled.
Will current :disabled condition leave it enabled?

rckopteri: Back to Draft - button is only shown when transport service is published.
It is inside 
(if published? )

PR #1069, opened 2019-06-25
166236387 matomo spa tracking

rckopteri:
# Added
* SPA tracking to matomo
* Production site-id to matomo


ote/src/cljs/ote/app/routes.cljs
```diff
@@ -154,11 +155,16 @@
                    (do
                      ;; Send startup events (if any) immediately after returning from this swap
                      (when (or event-leave event-to)
-                       (.setTimeout
-                         js/window
-                         (fn []
-                           (send-startup-events (vec (concat event-leave event-to))))
-                         0))
+                       (do
```
philosotractor: `do` not needed because it's already done implicitly by `when` :)
```suggestion
```
https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L493

PR #1073, opened 2019-06-26
Matomo url fix

jkorpelin:
# Fixed
* Removed tokens from urls sent to matomo.
* Show the "invalid token" notification first and when the token is checked show the "valid token" notification

### DIFF
--- a/ote/src/cljs/ote/app/routes.cljs
+++ b/ote/src/cljs/ote/app/routes.cljs
@@ -143,7 +143,12 @@
                      event-to (if (vector? event-to) event-to [event-to])
                      orig-app app
                      app (merge app navigation-data)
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/app/routes.cljs
```diff
@@ -143,7 +143,12 @@
                      event-to (if (vector? event-to) event-to [event-to])
                      orig-app app
                      app (merge app navigation-data)
-                     win-location (subs (str "/" (.. js/window -location -hash)) 1)]
+                     win-location (subs (.. js/window -location -hash) 1)
+                     win-location (if (boolean
```
philosotractor: This could be `if-let [page (#{:register :reset-password} (:page navigation-data)))`

PR #1076, opened 2019-06-27
Service search: fix view to use array_cat instead of coalesce

jkorpelin:
# Changed
* transport_service_search_result view to use create a concatenated list of arrays instead of finding the first non null value to select.


database/src/main/resources/db/migration/V1_154__fix_transport_service_search_view.sql
```diff
@@ -0,0 +1,27 @@
+-- View was using coalesce instead of creating a proper array of all the wanted data.
+CREATE OR REPLACE VIEW transport_service_search_result AS
+SELECT t.*, op.name as "operator-name", op."business-id" as "business-id",
+       -- Changed Start
```
rckopteri: 👍

PR #1083, opened 2019-07-09
166907227 email verification

jkorpelin:
# Added
* Email confirmation
   
# Fixed
* Login form returning status code 200 on every response. 

# Changed
* Users can no longer change their emails, because this would bring a lot of unwanted complexity to confirming email addresses. This feature could be done in a different ticket.
   
# Datamodel
* New table for email confirmation tokens
* User now has email-confirmed? column and confirmation timestamp column


ote/src/clj/ote/services/register.clj
```diff
@@ -25,50 +34,118 @@
   (if-not (valid-registration? form-data)
     ;; Check errors that should have been checked on the form
     {:success? false}
-    (with-transaction db
```
erno: transaction might still be needed? or is this called from inside an outer transaction?

PR #1093, opened 2019-07-24
167190723 admin edit user

jkorpelin:
# Added
* Ability for admins to edit users information

# Changed
* added a new users service and utils ns for encryptions


ote/src/cljs/ote/views/user_edit.cljs
```diff
@@ -0,0 +1,85 @@
+(ns ote.views.user-edit
```
rckopteri: should be in ote.views.admin. package

PR #1102, opened 2019-08-01
166440154 force gtfs download for service

rckopteri:
# Added
* admin panel: Added input field to enable gtfs package download for selected service

### DIFF
--- a/ote/src/clj/ote/tasks/gtfs.clj
+++ b/ote/src/clj/ote/tasks/gtfs.clj
@@ -18,7 +18,8 @@
             [ote.util.functor :refer [fmap]]
             [ote.util.collections :refer [map-by]]
             [ote.transit-changes.detection :as detection]
-            [ote.config.transit-changes-config :as config-tc])
+            [ote.config.transit-changes-config :as config-tc]
+            [ote.components.http :as http])
   (:import (org.joda.time DateTimeZone)))
 
 (defqueries "ote/tasks/gtfs.sql")
@@ -32,7 +33,12 @@
     "GTFS" :gtfs
     "Kalkati.net" :kalkati))
 
-(defn fetch-and-mark-gtfs-interface! [config db]
+(defn mark-gtfs-package-imported! [db gtfs-data]
+  (specql/update! db ::t-service/external-interface-description
+                  {::t-service/gtfs-imported (java.sql.Timestamp. (System/currentTimeMillis))}
+                  {::t-service/id (:id gtfs-data)})) ;; external-interface-description.id, not service id.
+
+(defn fetch-next-gtfs-interface! [config db]
   (tx/with-transaction
     db
     (let [blacklisted-operators {:blacklist (if (empty? (:no-gtfs-update-for-operators config))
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/app/controller/admin_transit_changes.cljs
```diff
@@ -93,12 +97,21 @@
                  {:on-success (tuck/send-async! ->SetSingleDetectionServiceId service-id)}))
     app))
 
+(define-event ForceInterfaceImportForGivenServiceSuccess [response]
+  {}
+  (assoc app :flash-message "GTFS paketti ladattu onnistuneesti"))
```
philosotractor: Ok, however the current flash-message is easy to miss.
If the result was embedded to the view, next to the button, using e.g. `ote.ui.common/generic-help` or `help` and styled blue like a few views already do, 
it would be persistent on the UI until user navigates away.

ote/src/clj/ote/tasks/gtfs.clj
```diff
@@ -32,34 +33,55 @@
...
+        ;; Load next gtfs package or package that is related to given service-id
+        (if (nil? service-id)
+          (fetch-next-gtfs-interface! config db)
+          (fetch-given-gtfs-interface! db service-id))
+        force-download? (integer? service-id)]
     (if (nil? gtfs-data)
-      (log/debug "No gtfs files to upload.")
+      (do
+        (log/debug "No gtfs files to upload.")
+        (http/transit-response "No gtfs files to upload." 400))
```
philosotractor: A 2xx status code should be used when an operation completed without errors.
In this case 200 or if we want to be exact, 204 No Content.

400 is not the best, because 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html
> 400 Bad Request
> The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

ote/src/clj/ote/tasks/gtfs.sql
```diff
@@ -11,6 +11,16 @@ SELECT eid.id as id, (eid."external-interface").url, eid.format[1], eid.license,
  ORDER BY "gtfs-imported" ASC LIMIT 1
    FOR UPDATE SKIP LOCKED;
 
+-- name: select-gtfs-url-for-service
+SELECT eid.id as id, (eid."external-interface").url, eid.format[1], eid.license,
+       ts."transport-operator-id" as "operator-id", ts.id as "ts-id", eid."gtfs-imported" as "last-import-date"
+  FROM "external-interface-description"  eid
```
philosotractor: An extra space char on this and next row between two words :)

ote/src/clj/ote/tasks/gtfs.clj
rckopteri: I added a notification box to inform users.

ote/src/clj/ote/tasks/gtfs.sql
rckopteri: fixed

PR #1103, opened 2019-08-01
167089456 authority email templates

rckopteri:
# Added
* operator-users: Add different emails for authority invites

### DIFF
--- a/ote/src/clj/ote/services/transport.clj
+++ b/ote/src/clj/ote/services/transport.clj
@@ -570,13 +570,15 @@
         _ (delete-old-tokens db ckan-group-id)]
     (http/transit-response users&invites 200)))
 
-(defn add-user-to-operator [email db new-member requester operator]
+(defn add-user-to-operator [email db new-member requester operator authority?]
   (let [member (create-member! db (:user_id new-member) (::t-operator/group-id operator))
         ;; Ensure that we do not send email to test user
         updated-member (if (= (:e2e-test-email (email-config/config)) (:user_email new-member))
                (assoc new-member :user_email (:e2e-test-amazon-simulator-email (email-config/config)))
                new-member)
-        title (str "Sinut on kutsuttu " (::t-operator/title operator) " -nimisen palveluntuottajan jäseneksi")
+        title (if authority?
+                (str "Olet saanut kutsun liittyä tarkastelemaan markkinaehtoisen henkilöliikenteen muutosilmoituksia NAP:ssa")
+                (str "Sinut on kutsuttu " (::t-operator/title operator) " -nimisen palveluntuottajan jäseneksi"))
         auditlog {::auditlog/event-type :add-member-to-operator
                   ::auditlog/event-attributes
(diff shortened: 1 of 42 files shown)


ote/src/clj/ote/services/transport.clj
```diff
@@ -757,17 +765,16 @@
         (authorization/with-group-check db user ckan-group-id
           #(operator-users-response db ckan-group-id)))
 
-      (POST "/transport-operator/:ckan-group-id/users"
-            {{:keys [ckan-group-id]}
+      (POST "/transport-operator/:ckan-group-id/:authority/users"
```
philosotractor: Because client cannot be relied upon on security matters, it seems dangerous for service to rely on client on if the group or user are authority-ones. 
- Service should  use the same API as for adding users to regular groups, and resolve if it's an authority group  internally (`users/transit-authority-group-id`).
Note also how /pre-notices/authority-group-id may be invoked by a non-authority client.
- The endpoint or `manage-adding-users-to-operator` should enforce that only authorized users are allowed to add users to the authority group.

ote/src/clj/ote/services/transport.clj
```diff
@@ -570,13 +570,15 @@
         _ (delete-old-tokens db ckan-group-id)]
     (http/transit-response users&invites 200)))
 
-(defn add-user-to-operator [email db new-member requester operator]
+(defn add-user-to-operator [email db new-member requester operator authority?]
```
philosotractor:
```suggestion
(defn add-user-to-group [email db new-member requester operator authority?]
```

ote/src/clj/ote/services/transport.clj
```diff
@@ -640,7 +648,7 @@
       (catch Exception e
         (log/warn (str "Error while inviting " user-email " ") e)))))
 
-(defn manage-adding-users-to-operator [email db requester operator form-data]
+(defn manage-adding-users-to-operator [email db requester operator form-data authority?]
```
philosotractor:
```suggestion
(defn manage-adding-users-to-group [email db requester operator form-data authority?]
```

ote/src/clj/ote/services/transport.clj
```diff
@@ -606,7 +610,7 @@
      :fullname (:user_name new-member)
      :email (:user_email new-member)}))
 
-(defn invite-new-user [email-config db requester operator user-email]
+(defn invite-new-user [email-config db requester operator user-email authority?]
```
philosotractor:
```suggestion
(defn invite-new-user-to-group [email-config db requester operator user-email authority?]
```

ote/src/clj/ote/services/transport.clj
rckopteri: I'm not sure if it is more dangerous because users can't invite other users to be part of operator was it a normal operator or authority.
But I changed the backend url back where it was and added db query to find out if operator is authority operator.

philosotractor: Okay, in that case...

PR #1104, opened 2019-08-01
166180208 remove username field 167647241 session to use userid not username

philosotractor:
# Fixed
* 6608475  services: users register fix for 167619111 future tokens deleted

# Changed
* test: fetch run-time the user id of username for session cookie
* db: user table add email field constaint unique not null
* views: admin, in admin user list replace username field using id field
* services: login, fetch user data based on id instead of username
* app: routes requires-authentication? check to use user-logged-in?
* views: front_page,main_header user-logged-in? moved to common controller
* views: main_header use appstate user key instead of username
* views:  remove username handling
* services: register-user to set db row id to username field
* services: save-user not to modify username db field
* controller: login, dissoc register from appstate when navigating away
* db: user username drop constraint unique not null

### DIFF
--- a/ote/test/clj/ote/integration/export/geojson_test.clj
+++ b/ote/test/clj/ote/integration/export/geojson_test.clj
@@ -3,7 +3,7 @@
             [clojure.test :as t :refer [use-fixtures deftest is testing]]
             [clojure.test.check.clojure-test :refer [defspec]]
             [clojure.test.check.properties :as prop]
-            [ote.test :refer [system-fixture *ote* http-post http-get sql-execute!]]
+            [ote.test :refer [system-fixture *ote* http-post http-get sql-execute! fetch-id-for-username]]
             [com.stuartsierra.component :as component]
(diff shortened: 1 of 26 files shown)


ote/test/clj/ote/integration/export/geojson_test.clj
```diff
@@ -36,7 +36,8 @@
   ;; Make it public so that GeoJSON export can return it
   (sql-execute! "UPDATE \"transport-service\" SET published = to_timestamp(0) where id = " id)
   (let [geojson-url (str "export/geojson/" transport-operator-id "/" id)
-        json-response (http-get "admin" geojson-url)]
+        json-response (http-get (fetch-id-for-username (:db ote.test/*ote*) "admin")
```
rckopteri: Just wondering if (fetch-id-for-username) could be done once with fixture :once

philosotractor: Good point, will do a PR about that.

PR #1108, opened 2019-08-05
167665859 table view rendering optimisation via react key optimisation

philosotractor:
# Changed
* 0d4a87414 ui: table component's react key optimisation by removing some indices
*  views: table-name key set for tables for several views

### DIFF
--- a/ote/src/cljs/ote/ui/table.cljs
+++ b/ote/src/cljs/ote/ui/table.cljs
@@ -19,9 +19,9 @@
                         :color "gray"}}
        opts])))
 
-(defn table [{:keys [height name->label key-fn label-style row-style
+(defn table [{:keys [table-name height name->label key-fn label-style row-style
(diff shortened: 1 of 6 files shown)


ote/src/cljs/ote/ui/table.cljs
```diff
@@ -78,9 +79,9 @@
         (doall
           (map-indexed
             (fn [i row]
-              ^{:key (if key-fn
-                       (key-fn row)
-                       (str i "-" random-table-id "-body-row"))}
+              ^{:key (str table-id "-b-row-" (if (and key-fn (key-fn row))
```
philosotractor: This `key-fn` handling could have been removed, based on testing views, but to avoid any risk of regression I left it still there. Just in case with some data it's better.

philosotractor:
@rckopteri  Could you please review this afterwards when you're able, thanks
Merging this for now to be able to continue.

PR #1109, opened 2019-08-06
167723828 clear deleted user email  167723801 clear email token on admins delete user

philosotractor:
# Fixed
* services: remove reqister.sql, the only query is also in user_service.sql
  services: register, use email-exists? form user_service.sql
* services: admin, clear email token when user is deleted
* services: allow delete-user! by setting email to id instead of null


philosotractor:
@rckopteri  Could you please review this afterwards when you're able, thanks
Merging this for now to be able to continue.

PR #1122, opened 2019-08-13
Css changes to fix overflow issues with ie 11

rckopteri:
# Fixed
* transit-visualization-page: Overflow issues with ie

### DIFF
--- a/ote/src/cljs/ote/style/transit_changes.cljs
+++ b/ote/src/cljs/ote/style/transit_changes.cljs
@@ -5,20 +5,14 @@
             [ote.theme.screen-sizes :refer [width-xxs width-xs width-sm width-md width-l width-xl]]))
 
 (def transit-changes-legend-container
-  {:align-items "flex-start"
-   :background-color "rgb(235,235,235)"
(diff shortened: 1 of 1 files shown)


ote/src/cljs/ote/style/transit_changes.cljs
```diff
@@ -5,20 +5,14 @@
...
    :padding "1rem"})
 
 (def transit-changes-icon-legend-row-container
-  {:align-items "center"
-   :display "flex"
-   :flex-direction "row"
-   :flex-wrap "wrap"
-   :justify-content "flex-start"})
+  {:display "flex"
+   :min-width "100%"
```
philosotractor: Why :min-width and not :width?

rckopteri: In this case it does not really matter. I only forced element to take all the width that is possible.

philosotractor: Allrighty, then `:width` should be preferred, in this case we want the element to fill it's parent container but not wider.

PR #1130, opened 2019-08-15
164534477 fix checkbox placement edge

rckopteri:
# Fixed
* Transit visualization page: Fixed next years calendar checkbox overflow issue with IE Edge
   
# Changed
*  Transit visualization page: Cryptic package download error text changed

### DIFF
--- a/ote/src/cljs/ote/views/transit_visualization/calendar.cljs
+++ b/ote/src/cljs/ote/views/transit_visualization/calendar.cljs
@@ -101,12 +101,15 @@
        [:div (stylefy/use-style (merge (style-base/flex-container "row")
                                        {:margin-top "1rem"
                                         :margin-bottom "1rem"}))
-        [ui/checkbox {:label "Näytä myös edellinen vuosi"
-                      :checked show-previous-year?
-                      :on-check #(e! (tv/->ToggleShowPreviousYear))}]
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/views/transit_visualization/calendar.cljs
```diff
@@ -101,12 +101,15 @@
        [:div (stylefy/use-style (merge (style-base/flex-container "row")
                                        {:margin-top "1rem"
                                         :margin-bottom "1rem"}))
-        [ui/checkbox {:label "Näytä myös edellinen vuosi"
-                      :checked show-previous-year?
-                      :on-check #(e! (tv/->ToggleShowPreviousYear))}]
-        [ui/checkbox {:label "Näytä myös seuraava vuosi"
-                      :checked show-next-year?
-                      :on-check #(e! (tv/->ToggleShowNextYear))}]]]
+        [:div {:style {:flex 1}}
```
philosotractor: Do we need a comment here "Width set via flex to avoid IE oveflowing element"? In case I touch this later :)
BTW Same could be accomplished using `width` property if flex was not involved.

rckopteri: No we don't need. :)

PR #1131, opened 2019-08-15
167727513 Simplify pre notice email subject to prevent issues in production

rckopteri:
# Fixed
* Simplify pre notice email subject to prevent issues in production

### DIFF
--- a/ote/src/clj/ote/services/admin.clj
+++ b/ote/src/clj/ote/services/admin.clj
@@ -38,7 +38,9 @@
             [clojure.spec.alpha :as spec]
             [ote.util.collections :as ote-coll]
             [ote.util.email-template :as email-template]
-            [ote.services.users :as srv-users]))
+            [ote.services.users :as srv-users]
+            [clj-time.format :as format])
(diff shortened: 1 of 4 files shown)


ote/src/clj/ote/services/admin.clj
```diff
@@ -509,6 +511,39 @@
...
+        _ (println "log-different-dates: maybe-working-subject = " maybe-working-subject)
+        date-str (str "cached-timezone " cached-timezone " /n "
+                      "current-t-now " current-t-now " /n "
+                      "timezone-now " timezone-now " /n "
+                      "problematic-formation " problematic-formation " /n "
+                      "working-formation " working-formation " /n "
+                      "problematic-subject " problematic-subject " /n "
+                      "maybe-working-subject " maybe-working-subject " /n ")]
+
+    (log/warn "Logging different date formations")
```
philosotractor: Should this log be before the let?

rckopteri: Yes it should, but it do not matter. :)

PR #1144, opened 2019-08-22
Admin view clear pre-notice sending result on view load

philosotractor:
# Changed
* views: admin: clear response of manual pre-notice sending on view load
* views: admin: simplify download single gtfs result notification hiccup

### DIFF
--- a/ote/src/cljs/ote/app/controller/admin_transit_changes.cljs
+++ b/ote/src/cljs/ote/app/controller/admin_transit_changes.cljs
@@ -14,6 +14,13 @@
             cljsjs.filesaverjs
             [ote.app.routes :as routes]))
 
+(define-event InitAdminDetectedChanges []
+  {}
+  (update-in app
+             [:admin :pre-notice]
+             dissoc
+             :pre-notice-notify-send))
+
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/app/controller/admin_transit_changes.cljs
```diff
@@ -14,6 +14,13 @@
             cljsjs.filesaverjs
             [ote.app.routes :as routes]))
 
+(define-event InitAdminDetectedChanges []
```
rckopteri: You could use on-navigate event

philosotractor: True, but that doesn't happen when user is on the page and does something which causes a page reload. 
Like pushes one of the other action buttons to trigger a completely different backend call.
In which case the old result boxes should be cleared.

BTW I first put also the single gtfs package download result clearing into this same event, but reverted it after not getting it working perfectly out-of-the box. 
Apparently completion of that causes a page redraw which cleared also the new result, so either that redraw has to be made differently or we skip clearing it's result for now. I chose the latter.

PR #1145, opened 2019-08-23
167283447 update service viewer

rckopteri:
# Added
* Service-viewer: Added support for terminal, parking and rental services


ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -493,32 +531,362 @@
...
+     [spacer]]))
+
 (defn service-view
   [e! {{to :transport-operator ts :transport-service} :service-view}]
-  (let [interfaces (::t-service/external-interfaces ts)
-        warnings (get-in ts [::t-service/passenger-transportation ::t-service/luggage-restrictions])
-        real-time-info-data (get-in ts [::t-service/passenger-transportation ::t-service/real-time-information])
-        pre-booking-data (get-in ts [::t-service/passenger-transportation ::t-service/advance-reservation])
-        booking-data (get-in ts [::t-service/passenger-transportation ::t-service/booking-service])
+  (let [service-sub-type (get ts ::t-service/sub-type)
```
philosotractor: `let` could be put inside the `if`so that all the bindings are skipped when they're not needed.

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -493,32 +531,362 @@
...
-       [real-time-info (tr [:service-viewer :real-time-info]) real-time-info-data]
-       [pre-booking (tr [:service-viewer :advance-reservation]) pre-booking-data]
-       [booking-service (tr [:service-viewer :reservation-service]) booking-data]
-       [accessibility-and-other-services (tr [:service-viewer :accessibility-and-other-services]) accessibility-data]
-       [price-information (tr [:service-viewer :price-information]) pricing-data]
-       [service-hours (tr [:service-viewer :service-hours]) service-hours-data]])))
+      (case service-sub-type
+        :rentals
+        [:div
+         [service-header (::t-operator/name to) (::t-operator/id to) (::t-service/id ts)]
```
philosotractor: Having identical definition for a sub-component under each case, one alternative is to declare a local contructor function in the let and call that under each case. That way any updates to the sub-component creation can be done in one place.

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -493,32 +531,362 @@
...
+                                  (not (nil? (:minutes data)))
+                                  (> (:minutes data) 0)) :minutes
+                                (and
+                                  (not (nil? (:hours data)))
+                                  (> (:hours data) 0)) :hours
+                                (and
+                                  (not (nil? (:days data)))
+                                  (> (:days data) 0)) :days
+                                :default :hours)
+        restriction-value (restriction-time-type data)]
```
philosotractor: Isn't `restrisction-time-type` a symbol which gets bound in `let` initialisation, so instead of `restriction-value` code could refer directly `restrisction-time-type`.
(Or if not, then `restrisction-time-type?` should be declared as `fn`)

https://clojuredocs.org/clojure.core/some_q

philosotractor: BTW `(some? v)` is equals to `(not (empty? v)`

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -493,32 +531,362 @@
...
+                ^{:key (str vehicle-type "-" license-required "-" price)}
+                [common-ui/information-row-with-option (tr [:field-labels :rentals :ote.db.transport-service/price-classes]) (str price " €") false]))]
+           [:div
+            [common-ui/information-row-with-option (tr [:field-labels :rentals :ote.db.transport-service/minimum-age]) minimum-age false]
+            (doall
+              (for [price-row (::t-service/price-classes vc)
+                    :let [unit (::t-service/unit price-row)]]
+                ^{:key (str vehicle-type "-" minimum-age "-" unit)}
+                [common-ui/information-row-with-option (tr [:service-viewer :pricing-basis]) unit false]))]]))])
+   [spacer]])
```
philosotractor: If the spacer is not involved to this sub-component, but meant to separate two different sub-components, then spacer should be moved one level up outside this sub-component into the `case`.
Reasoning: if this is re-used, a spacer is not necessarily appropriate there.

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -493,32 +531,362 @@
...
+            (doall
+              (for [price-row (::t-service/price-classes vc)
+                    :let [price (::t-service/price-per-unit price-row)]]
+                ^{:key (str vehicle-type "-" license-required "-" price)}
+                [common-ui/information-row-with-option (tr [:field-labels :rentals :ote.db.transport-service/price-classes]) (str price " €") false]))]
+           [:div
+            [common-ui/information-row-with-option (tr [:field-labels :rentals :ote.db.transport-service/minimum-age]) minimum-age false]
+            (doall
+              (for [price-row (::t-service/price-classes vc)
+                    :let [unit (::t-service/unit price-row)]]
```
philosotractor: `:let unit` could be removed and define it in the :key call

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -402,6 +405,50 @@
         true]]]
      [spacer]]))
 
+(defn- rental-accessibility-and-other-services
+  [title data]
+  (let [url (:url data)
```
philosotractor: Not an issue, but `url` could be destructured in the function args list.

ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -19,11 +20,17 @@
...
             [ote.app.controller.place-search :as place-search]
             [ote.ui.form-fields :as form-fields]
-            [ote.style.base :as base]))
+            [ote.style.base :as base]
+            [clojure.string :as str]))
 
 (defonce shown-language
          (r/atom (string/upper-case (name @selected-language))))
 
+(defn format-time-key->str [time-map first-key second-key]
```
philosotractor:
```suggestion
(defn format-time-key->str [time-map first-key second-key]
  (let [v (str (get-in time-map [first-key second-key]))]
    (str (when (= (count v) 1)
           "0")
         v)))
```

ote/src/cljs/ote/views/rental.cljs
```diff
@@ -243,18 +248,18 @@
...
      {:label (tr [:passenger-transportation-page :header-pick-up-locations])
-      :columns 3}
+      :columns 1}
 
      {:name ::t-service/pick-up-locations
       :type :table
       :prepare-for-save values/without-empty-rows
       :table-fields [{:name ::t-service/pick-up-name
-                      :width "20%"
+                      ;:width "20%"
```
philosotractor: Commented widths could be removed

ote/src/cljs/ote/views/service_viewer.cljs
rckopteri: This is a good idea.
I moved identical components before the case and leave only relevant components.

ote/src/cljs/ote/views/service_viewer.cljs
rckopteri: Spacer was used this way in all sub-components in the original design.
I decided not to refactor this design in this pr. (I ended up to fix other pages which is not wise)

ote/src/cljs/ote/views/service_viewer.cljs
rckopteri: Correct, but I leave it as it is.

ote/src/cljs/ote/views/service_viewer.cljs
rckopteri: Correct, but I leave it as it is.

ote/src/cljs/ote/views/service_viewer.cljs
rckopteri: Nice. Thanx. This is way better.

ote/src/cljs/ote/views/rental.cljs
rckopteri: removed

PR #1161, opened 2019-09-10
168311599 reorder removed stops

rckopteri:
# Fixed
* transit visualization page: Reorder removed stops by adding additional sorting function

### DIFF
--- a/ote/src/cljc/ote/transit_changes.cljc
+++ b/ote/src/cljc/ote/transit_changes.cljc
@@ -138,25 +138,13 @@
 (defn earliest-stop-sequence [stop]
   (:gtfs/stop-sequence stop))
 
-(defn interval->seconds
-  "Converts interval to seconds. Uses hard coded days of year and days of month because it is accurate enough. Interval
-  doesn't know which year is currently and it doesn't care."
-  [{:keys [years months days hours minutes seconds] :as i}]
-  (+
-    (* (or years 0) 365 24 60 60)
-    (* (or months 0) 31 24 60 60)
-    (* (or days 0) 24 60 60)
-    (* (or hours 0) 60 60)
-    (* (or minutes 0) 60)
-    (or seconds 0)))
-
 (defn smaller-sequence-for-stop [stop stops]
   (let [date1-departure-time (:gtfs/departure-time-date1 stop)
         reversed-stops (reverse stops)
         first-smaller-sequence (keep
                                  (fn [s]
                                    (let [sequence (:gtfs/stop-sequence s)]
-                                     (if (> (interval->seconds (:gtfs/departure-time-date1 s)) (interval->seconds date1-departure-time))
+                                     (if (> (time/interval->seconds (:gtfs/departure-time-date1 s)) (time/interval->seconds date1-departure-time))
                                        (assoc stop :gtfs/stop-sequence (- sequence 0.1))
                                        nil)))
                                  reversed-stops)]
--- a/ote/src/cljc/ote/transit_changes.cljc
+++ b/ote/src/cljc/ote/transit_changes.cljc
@@ -138,6 +138,48 @@
 (defn earliest-stop-sequence [stop]
   (:gtfs/stop-sequence stop))
 
+(defn interval->seconds
+  "Converts interval to seconds. Uses hard coded days of year and days of month because it is accurate enough. Interval
+  doesn't know which year is currently and it doesn't care."
+  [{:keys [years months days hours minutes seconds] :as i}]
+  (+
+    (* (or years 0) 365 24 60 60)
+    (* (or months 0) 31 24 60 60)
+    (* (or days 0) 24 60 60)
+    (* (or hours 0) 60 60)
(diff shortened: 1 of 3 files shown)


ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -138,6 +138,48 @@
 (defn earliest-stop-sequence [stop]
   (:gtfs/stop-sequence stop))
 
+(defn interval->seconds
```
philosotractor: This should be put into time.cljc - coincidentally I accidentally did it in https://github.com/finnishtransportagency/mmtis-national-access-point/pull/1159 :)

rckopteri: Merge your pr and I'll update my pr to use that fn.

ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -138,6 +138,36 @@
 (defn earliest-stop-sequence [stop]
   (:gtfs/stop-sequence stop))
 
+(defn smaller-sequence-for-stop [stop stops]
+  (let [date1-departure-time (:gtfs/departure-time-date1 stop)
+        reversed-stops (reverse stops)
+        first-smaller-sequence (keep
+                                 (fn [s]
+                                   (let [sequence (:gtfs/stop-sequence s)]
+                                     (if (> (time/interval->seconds (:gtfs/departure-time-date1 s)) (time/interval->seconds date1-departure-time))
```
philosotractor:
```suggestion
                                     (when (> (time/interval->seconds (:gtfs/departure-time-date1 s)) (time/interval->seconds date1-departure-time))
```

ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -138,6 +138,36 @@
...
+                                       (assoc stop :gtfs/stop-sequence (- sequence 0.1))
+                                       nil)))
```
philosotractor:
```suggestion
                                       )))
```

philosotractor:
If I got this right, stops are 
first ordered by `juxt earliest-departure-time earliest-stop-sequence`
then removed stops without `:gtfs/departure-time-date2` are moved earlier in the list based on `:gtfs/departure-time-date1` values of stops (by `smaller-sequence-for-stop`).
This is done only with regards stops before the removed stop.
So this solves cases where a removed stop is "too down/late" in the list but not if it's too "up/early".

Would following approach handle all cases, where removed stops should be sorted based on `:gtfs/departure-time-date1`?
- take a removed stop
- iterate all stops starting from end (reversed)
- compare removed stop to each stop, 
find first stop in reversed list having smaller`:gtfs/departure-time-date1` (D1) 
- set removed stop's index to greater than that and break loop
Pseudocode:
```
if remstop-date1 > stop-date1
then remstop-ix = stop-ix + 0.1
break
```

ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -138,6 +138,39 @@
...
+                                 (fn [s]
+                                   (let [sequence (:gtfs/stop-sequence s)]
+                                     (when (< (time/interval->seconds (:gtfs/departure-time-date1 s)) (time/interval->seconds date1-departure-time))
+                                       (assoc stop :gtfs/stop-sequence (* sequence 1.01))))) ; create new sequence nro for removed stop
+                                 reversed-stops)]
+    (if (or (empty? modified-stop) (nil? modified-stop))
+      stop
+      modified-stop)))
+
+(defn reorder-removed-stops
```
philosotractor: Would following simplification work?

- set following as inititial `val` (https://clojuredocs.org/clojure.core/reduce) of reduce:
```
(filterv #(not (nil? (:gtfs/departure-time-date2 %))) indexed-stops)
``` 
- in reduce return 
```
(conj result (sequence-for-stop stop result))
```

philosotractor: + sort-by to conj result

rckopteri: Not sure if I want make it more compact. It is easier and faster to read and understand when things are stored in named variables. But I'm not sure if those names are that good :)

philosotractor: a) No problem on documenting by using intermediate symbols with good names
(though one school of thought e.g. in "Elements of Clojure" is that if once can't come up with a good name one should a) sleep on it or b) not use any :) )

b) You might have missed the key point I was proposing: having `reduce` act as a function and using _only_ symbols which are given to it as arguments.
Now it's using ` other-stops` outside the reduce context.
It would me more maintainable and simple if ` other-stops` is given as the initial value 
and reduce iteration conjoins the added stop into it and returns the result for next iteration,
which does the same again.
:)

philosotractor: That will remove one `if` in the reduce. And actually the `if` shouldn't be necessary: `conj coll nil` returns coll so it's safe to conjoin nil.

PR #1166, opened 2019-09-11
168367747 dynamic db feature flag for sea route maintenance mode

philosotractor:
# Added
* language: error landing page and maintenance break strings
* services
routes: :maintenance-break-sea-route flag checked from db
routes: endpoints wrapped to service state check handler
* db
cljc specql namespace for feature variation
create feature_variation table and feature_type
* views
route-wizard: load stops failure handler to common servererror
route-list: load routes failure handler to common servererror
main: error-landing page selection
error-landing: first WIP version
 ote.sovellus height to 100vh to fill screen and put footer down
* controller: common: ServerError to redirect 503 to error-landing page
* ote: revert lein-figwheel to 0.5.13, connecting to browser not ok

### DIFF
--- a/ote/src/clj/ote/services/routes.clj
+++ b/ote/src/clj/ote/services/routes.clj
@@ -19,7 +19,8 @@
             [cheshire.core :as cheshire]
             [ote.db.tx :as tx]
             [jeesql.core :refer [defqueries]]
-            [ote.environment :as environment])
+            [ote.environment :as environment]
+            [ote.db.feature])                               ; specql table definitions
   (:import (org.postgis PGgeometry Point Geometry)))
 
 (defqueries "ote/services/routes.sql")
--- a/ote/src/clj/ote/services/routes.clj
+++ b/ote/src/clj/ote/services/routes.clj
@@ -25,13 +25,14 @@
 
 (defqueries "ote/services/routes.sql")
 
-(defn- service-state-response [db]
-  (when (:ote.db.feature/value
-          (first
-            (fetch db
-                   :ote.db.feature/feature-variation
-                   (specql/columns :ote.db.feature/feature-variation)
-                   {:ote.db.feature/feature :maintenance-break-sea-route})))
+(defn- service-state-response [db user]
+  (when (and (not (:admin? user))
+             (:ote.db.feature/value
+               (first
+                 (fetch db
+                        :ote.db.feature/feature-variation
+                        (specql/columns :ote.db.feature/feature-variation)
+                        {:ote.db.feature/feature :maintenance-break-sea-route}))))
     (http/no-cache-transit-response "Under maintenance" 503)))
 
 (defn- interface-url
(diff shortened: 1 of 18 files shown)


philosotractor:
Probably `no-cache-transit-response` should be used used in route endpoint, will improve that in following PR

ote/src/clj/ote/services/routes.clj
```diff
@@ -19,11 +19,21 @@
...
             [ote.db.tx :as tx]
             [jeesql.core :refer [defqueries]]
-            [ote.environment :as environment])
+            [ote.environment :as environment]
+            [ote.db.feature])                               ; specql table definitions
   (:import (org.postgis PGgeometry Point Geometry)))
 
 (defqueries "ote/services/routes.sql")
 
+(defn- service-state-response [db]
```
rckopteri: It is private function and only related to this service but still wondering if it should be renamed something like 'valid-route-service-state?'

ote/src/cljc/ote/db/feature.cljc
```diff
@@ -0,0 +1,19 @@
...
+  "Data model for dynamic feature flags"
+  (:require
+            #?@(:clj [[ote.db.specql-db :refer [define-tables]]
+                      [specql.postgis]
+                      [clj-time.coerce :as tc]]
+                :cljs [[cljs-time.coerce :as tc]])
+            [specql.impl.registry]
+            [specql.data-types]
+            [ote.db.common]
+            ;[ote.db.modification]
```
rckopteri: comment

ote/src/cljs/ote/views/error_landing.cljs
```diff
@@ -0,0 +1,15 @@
+(ns ote.views.error-landing
```
rckopteri: Consider moving to own package like errors or somethin. ote.views.errors.error-landing

ote/src/clj/ote/services/routes.clj
philosotractor: I see what your going for. However,
Logic is naming is the same as in admin service side: function returns a request response, hence name should have "-response" suffix. "get-route-response" in the file follows the same pattern.
Suffix "?" is for boolean return values.

ote/src/cljs/ote/views/error_landing.cljs
philosotractor: Ok...what role and purpose that package would serve, in addition to the error landing page?

rckopteri: We have quite many files in the views folder/package. We should split them to sub folders to help us navigate and find relevant files.

e.g. we have 2 at least 5 files that relates to transport-service and they are in alphabetical order. If they were in single sub folder finding them would be faster.

And from this point of view I'm expecting that we will have more than one error page in a few months so I would suggest that we add one folder for those :)

philosotractor: As discussed, design is to have just one reused error_landing view and file. Nevertheless I moved this to new package.

PR #1168, opened 2019-09-12
168453305 - sea route admin tab - some routes missing

rckopteri:
# Fixed
* Admin sea route tab: Fixed sql query for routes which are missing calendar rules.

### DIFF
--- a/ote/src/clj/ote/services/admin.sql
+++ b/ote/src/clj/ote/services/admin.sql
@@ -122,7 +122,7 @@ SELECT
        r.created AS "created"
   FROM
        "transit_route" r,
-       lateral unnest(
+       LATERAL unnest(
            CASE WHEN array_length(r."service-calendars", 1) >= 1
                     THEN r."service-calendars"
                 ELSE '{null}'::transit_service_calendar[]
--- a/ote/src/clj/ote/services/admin.sql
+++ b/ote/src/clj/ote/services/admin.sql
@@ -121,7 +121,12 @@ SELECT
        r.modified AS "modified",
        r.created AS "created"
   FROM
-       "transit_route" r, unnest(r."service-calendars") c, unnest(c."service-rules") ru,
+       "transit_route" r,
+       lateral unnest(r."service-calendars") c,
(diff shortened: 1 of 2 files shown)


ote/src/cljs/ote/views/admin/sea_routes.cljs
```diff
@@ -51,7 +51,8 @@
       3 ::transit/wednesday
       4 ::transit/thursday
       5 ::transit/friday
-      6 ::transit/saturday)))
+      6 ::transit/saturday
+      ::transit/sunday)))
```
philosotractor: Didn't go deep into this logic, so just double-checking is defaulting to sunday ok in all cases? It was already handled as index `0`?

ote/src/clj/ote/services/admin.sql
```diff
@@ -121,7 +121,17 @@ SELECT
        r.modified AS "modified",
        r.created AS "created"
   FROM
-       "transit_route" r, unnest(r."service-calendars") c, unnest(c."service-rules") ru,
+       "transit_route" r,
+       lateral unnest(
```
philosotractor:
```suggestion
       LATERAL unnest(
```

ote/src/clj/ote/services/admin.sql
```diff
@@ -121,7 +121,17 @@ SELECT
...
+           CASE WHEN array_length(r."service-calendars", 1) >= 1
+                    THEN r."service-calendars"
+                ELSE '{null}'::transit_service_calendar[]
+           END) c,
+       lateral unnest (
```
philosotractor:
```suggestion
       LATERAL unnest (
```

ote/src/cljs/ote/views/admin/sea_routes.cljs
rckopteri: Yes, it is ok.

If service calendar and service rule is not given we don't have any dates when route is supposed to be active (ajopäivät, don't know what it is in english). This prevents errors when ajopäivät is not given.

PR #1169, opened 2019-09-13
168453323 default operator admin

rckopteri:
# Changed
* sea route editor: Show operator data from :selected-operator key

### DIFF
--- a/ote/src/cljs/ote/app/controller/own_services.cljs
+++ b/ote/src/cljs/ote/app/controller/own_services.cljs
@@ -5,7 +5,8 @@
             [ote.localization :as localization :refer [tr]]
             [ote.app.controller.common :refer [->ServerError]]
             [ote.util.url :as url-util]
-            [ote.db.transport-operator :as t-operator]))
+            [ote.db.transport-operator :as t-operator]
+            [ote.app.routes :as routes]))
 
 (define-event SearchSuccess [result]
               {}
@@ -84,3 +85,13 @@
                   {}
                   {:on-success (tuck/send-async! ->RemoveSelectionSuccess transport-operator-id service-id)})
                 app))
+
+; Ensure that :transport-operator is not nil
+(define-event VerifyOperatorSelection []
+  {}
+  (if (nil? (:transport-operator app))
+    (assoc app :transport-operator (:transport-operator (first (get app :transport-operators-with-services))))
+    app))
+
+(defmethod routes/on-navigate-event :own-services [_]
+  (->VerifyOperatorSelection))
--- a/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
+++ b/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
@@ -38,7 +38,8 @@
     (-> app
         (assoc-in [:route :step] :basic-info)
         (assoc-in [:route ::transit/route-type] :ferry)
-        (assoc-in [:route ::transit/transport-operator-id] (get-in app [:transport-operator ::t-operator/id])))))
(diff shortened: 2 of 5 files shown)


ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -301,15 +307,20 @@
...
-      (-> app
-          (assoc :route response)
-          ;; make sure we don't overwrite loaded stops
-          (assoc-in [:route :stops] (get-in app [:route :stops]))
-          (assoc-in [:route ::transit/stops] stops)
-          (assoc-in [:route ::transit/trips] trips)
-          (assoc-in [:route ::transit/service-calendars] service-calendars))))
+          trips (vec (map-indexed (fn [i trip] (assoc trip ::transit/service-calendar-idx i)) trips))
+          operator-id (::transit/transport-operator-id response)]
+      (do
```
philosotractor:
```suggestion
```

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -285,6 +287,10 @@
   (process-event [{response :response} app]
     (assoc-in app [:route :stops] response))
 
+  LoadRouteOperatorResponse
```
philosotractor: While route operator is being loaded, will UI display potentially incorrect data containing whatever was left into appstate by previous visit to this view?
So the sea route operator section should display a loading spinner while operator data is not available
(or alternatively block the whole view)

ote/src/cljs/ote/app/controller/own_services.cljs
```diff
@@ -83,4 +84,14 @@
                        (url-util/encode-url-component transport-operator-id))
                   {}
                   {:on-success (tuck/send-async! ->RemoveSelectionSuccess transport-operator-id service-id)})
-                app))
\ No newline at end of file
+                app))
+
+; Ensure that :transport-operator is not nil
```
philosotractor:
```suggestion
;; Ensure that :transport-operator is not nil
```

philosotractor: Could this comment what case is this check fixing? A bit surprising to see `app` can be nil.

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
rckopteri: I was adding the spinner but decided to test it first and figured that we don't need the spinner because response is so quick.
But maybe we should add it anyway?

ote/src/cljs/ote/app/controller/own_services.cljs
philosotractor: Instead of `VerifyOperatorSelection` should the event be `InitOwnServices` or something similar?
If we had an event naming convention, it could be "Init<view or component name>"

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
philosotractor: I think we should. It's normally fast, but as we've seen in some cases a request may take long using a slow device running low on memory and with corporate sw slowing the browser down. Or with occasional network glitch. When using a loading placeholder, it's guaranteed we don't need to get back to this.
A slow request which blocks user is not as critical, but here user may end up being displayed wrong data.

rckopteri: Added two spinner. For the route and for the operator.

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
rckopteri: removed

PR #1172, opened 2019-09-17
164826821 random fixes

rckopteri:
# Fixed
* Sea route on-blur when save is clicked without clicking "off" from homepage input field
* Admin sea-route-list last active day when only added calendar date is given and no calendar rule

### DIFF
--- a/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
+++ b/ote/src/cljs/ote/app/controller/route/route_wizard.cljs
@@ -339,6 +339,8 @@
   UpdateOperatorHomepageResponse
   (process-event [{response :response} app]
     (-> app
+        ; clear operator homepage from route - to prevent unnecessary save
+        (dissoc [:route :operator-homepage])
         (update :transport-operators-with-services
                 (fn [operators]
                   (map
@@ -367,7 +369,9 @@
   (process-event [{new-homepage :new-homepage} app]
     (-> app
         (assoc-in [:selected-operator ::t-operator/homepage] new-homepage)
-        (assoc-in [:transport-operator ::t-operator/homepage] new-homepage)))
+        (assoc-in [:transport-operator ::t-operator/homepage] new-homepage)
+        ; Update operator homepage to route while operator homepage is not saved to db
+        (assoc-in [:route :operator-homepage] new-homepage)))
 
   AddStop
   (process-event [{feature :feature} app]
--- a/ote/src/cljs/ote/views/admin/sea_routes.cljs
+++ b/ote/src/cljs/ote/views/admin/sea_routes.cljs
@@ -64,7 +64,9 @@
                                (if (true? ((weekday-index-to-keyword last-weekday-index) route))
                                  (* -1 index)
                                  (recur (inc index) (dec last-weekday-index)))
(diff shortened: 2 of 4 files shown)


ote/src/cljs/ote/views/admin/sea_routes.cljs
```diff
@@ -64,7 +64,9 @@
                                (if (true? ((weekday-index-to-keyword last-weekday-index) route))
                                  (* -1 index)
                                  (recur (inc index) (dec last-weekday-index)))
-                               nil))]
+                               ; return 0 days only if to-date and weekday are given.
+                               (when (and to-date weekday)
+                                 0)))]
     (if (not (nil? last-weekday-count))
```
philosotractor:
```suggestion
(when last-weekday-count
      (time/native->date
        (t/plus (time/native->date-time to-date) (t/days last-weekday-count))))
```

ote/src/cljs/ote/views/admin/sea_routes.cljs
```diff
@@ -64,7 +64,9 @@
                                (if (true? ((weekday-index-to-keyword last-weekday-index) route))
                                  (* -1 index)
                                  (recur (inc index) (dec last-weekday-index)))
-                               nil))]
+                               ; return 0 days only if to-date and weekday are given.
```
philosotractor:
```suggestion
                               ;; return 0 days only if to-date and weekday are given.
```
https://guide.clojure.style/#two-semicolons-for-code-fragment

+Add a "why" into the comment if reasonable.

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
```diff
@@ -339,6 +339,8 @@
   UpdateOperatorHomepageResponse
   (process-event [{response :response} app]
     (-> app
+        ; clear operator homepage from route - to prevent unnecessary save
+        (dissoc [:route :operator-homepage])
```
philosotractor: Or rather select-keys (specql/columns <tablename> c) in service side.

ote/src/cljs/ote/views/admin/sea_routes.cljs
rckopteri: Replaced with when-not

ote/src/cljs/ote/views/admin/sea_routes.cljs
rckopteri: why added

ote/src/cljs/ote/app/controller/route/route_wizard.cljs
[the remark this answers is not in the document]
rckopteri: Maybe in distant future. Actually :operator-homepage key is removed in every case, so it won't cause problems but it is not a robust way to handle things.
We should refactor all our form-data codes in same time.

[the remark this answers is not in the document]
philosotractor: While this of course does not block this PR,
I might disagree on the reasoning. Adding such a select-keys to service side is a simple and safe one-liner and can be implemented incrementally and does not require any other refactoring :)

PR #1174, opened 2019-09-18
165239455 ui improvements

rckopteri:
# Fixed
* Reduce header element bottom padding
* Fix invite users to operator sql query
* Add new small-text css element
* Remove black bottom borders from table component
* Remove black bottom borders from route list tables and own services tables

### DIFF
--- a/ote/src/cljs/ote/ui/table.cljs
+++ b/ote/src/cljs/ote/ui/table.cljs
@@ -28,7 +28,6 @@
         table-row-hover-color colors/gray400]
     [ui/table (merge
                 {:wrapperStyle {:overflow "visible"}
-                 :header-style {:border (str "solid 3px " colors/gray400)}
                  :body-style {:padding "3px"}
                  ;; FIXME: When we have tooltips in header labels, body does not need overflow: visible.
                  ;;        But, if any row includes tooltips, the body must also have visible overflow.
@@ -48,11 +47,14 @@
                 (when class
                   {:class class}))
      [ui/table-header {:key (str table-id "-hdr")
-                       :style {:overflow "visible" :background-color (color :grey300)}
+                       :style {:overflow "visible"
+                               :background-color (color :grey300)
+                               :border-bottom (str "1px solid " colors/gray650 " !important")}
                        :adjust-for-checkbox false
                        :display-select-all false}
       [ui/table-row {:key (str table-id "-hdr-row")         ;; When adding more rows remember to add their id to key
-                     :style {:overflow "visible"}
+                     :style {:overflow "visible"
+                             :border-bottom (str "1px solid " colors/gray650 " !important")}
                      :selectable false}
        (doall
          (map-indexed
@@ -87,6 +89,7 @@
                               (stylefy/use-style (merge {:background-color (if (even? i)
                                                                              table-row-color
                                                                              table-row-color-alt)} ; Add stripes
+                                                        {:border-bottom (str "1px solid " colors/gray650 " !important")}
                                                         (when on-select
                                                           {:cursor "pointer"
                                                            :transition (str "background-color " transitions/fast-ease-in-out)
--- a/ote/src/cljs/ote/views/route/route_list.cljs
+++ b/ote/src/cljs/ote/views/route/route_list.cljs
@@ -32,11 +32,12 @@
                                 :keys [show-delete-modal?]
                                 :as route}]
   [:span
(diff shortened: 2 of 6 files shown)


ote/src/cljs/ote/ui/table.cljs
```diff
@@ -48,11 +47,14 @@
                 (when class
                   {:class class}))
      [ui/table-header {:key (str table-id "-hdr")
-                       :style {:overflow "visible" :background-color (color :grey300)}
+                       :style {:overflow "visible"
+                               :background-color (color :grey300)
+                               :border-bottom (str "1px solid " colors/gray650 " !important")}
```
philosotractor: Why only -bottom? Should it be all borders?

ote/src/cljs/ote/ui/table.cljs
```diff
@@ -48,11 +47,14 @@
...
                        :adjust-for-checkbox false
                        :display-select-all false}
       [ui/table-row {:key (str table-id "-hdr-row")         ;; When adding more rows remember to add their id to key
-                     :style {:overflow "visible"}
+                     :style {:overflow "visible"
+                             :border-bottom (str "1px solid " colors/gray650 " !important")}
```
philosotractor: Why only -bottom? Should it be all borders?

ote/src/cljs/ote/ui/table.cljs
rckopteri: Nope. Issue was only with bottom-border

ote/src/cljs/ote/views/route/route_list.cljs
```diff
@@ -79,19 +80,21 @@
...
                                (.preventDefault %)
                                (e! (fp/->ChangePage :edit-route {:id route-id})))} (t-service/localized-text-with-fallback @selected-language name)]]
             [ui/table-row-column {:style {:width "15%"}} (time/format-timestamp-for-ui modified)]
             [ui/table-row-column {:style {:width "15%"}} (time/format-timestamp-for-ui created)]
-            [ui/table-row-column {:style {:width "13%"}}
-             [ui/icon-button {:href "#"
-                              :on-click #(do
-                                           (.preventDefault %)
-                                           (e! (fp/->ChangePage :edit-route {:id route-id})))}
+            [ui/table-row-column {:style {:width "11%" :padding-left "10px"}}
```
philosotractor: padding should probably be rem or em

ote/src/cljs/ote/views/route/route_list.cljs
```diff
@@ -134,16 +145,24 @@
...
-                                   (tr [:route-list-page :no-draft-routes]))]]
-   [:h4 (tr [:route-list-page :header-public-routes])]
-   [:p (tr [:route-list-page :desc-public-routes])]
-   [:div.public
-    [route-table e! public-routes (when (empty? public-routes)
-                                    (tr [:route-list-page :no-public-routes]))]]])
+   [:div {:style {:padding-bottom "1rem"}}
+    [:h4 (tr [:route-list-page :header-route-drafts])]
+    [:p {:style (merge
+                  {:margin-top "0px"}
```
philosotractor: rem?

ote/src/cljs/ote/views/route/route_list.cljs
rckopteri: Padding is needed because it moves action icons to left. Icons are sized as px, so px is needed here also.

ote/src/cljs/ote/ui/table.cljs
rckopteri: Border is set for multiple elements but still I think that we wouldn't win anything by adding it to def and merging styles here. 
This border issue comes from material ui. And material-ui seems to add border to table-row item or table-row AND table-header item. So I had to overwrite them in both elements.

philosotractor: Yes, the need for setting the style for all these components is not contested here.

Point was duplicate code. 
Duplication leads to someone potentially changing only some/one of the locations in future and missing some. That That could be avoided by 
1) Using a def'd style makes it explicit and developer-proof
2) Using css `inherit` also removes the duplication because in this case child component is meant to follow parent component style without differing from it. 
E.g. On browser dev console setting the `tr` border to "inherit" fixes the row and all it's td's without the need to change the td styles at all.

There wouldn't be any extra effort or complexity on above alternativecs, on the contrary they would be more future-proof and simple.
But okay I'll let this go :)

PR #1180, opened 2019-09-23
163385823 foreign postal code

rckopteri:
# Added
* Added support for foreign postal codes (was char 5 is now varchar 10)

### DIFF
--- a/database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
+++ b/database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
@@ -1,6 +1,8 @@
 ALTER TYPE address
     ADD ATTRIBUTE "foreign_postal_code" VARCHAR(10);
 
+-- Create backup for "transport-service" table
+CREATE TABLE "transport-service_backup" AS TABLE "transport-service";
 
 CREATE OR REPLACE FUNCTION convert_transport_service_address_postalcodes()
 RETURNS VOID AS $$
@@ -13,12 +15,22 @@ BEGIN
               WHERE service.id = ts.id);
 
     -- update rental pick up addressess
-    UPDATE "transport-service" "t-service"
-       SET (rentals."pick-up-locations") =
-            (SELECT array_agg(locs.*)
-               FROM "transport-service" service,
-                    LATERAL UNNEST((service.rentals)."pick-up-locations") locs
-              WHERE service.id = "t-service".id);
+    UPDATE "transport-service" as ts
+    SET (rentals."pick-up-locations") =
+            (SELECT array_agg(ROW(pa."pick-up-name",
+                                  pa."pick-up-type",
+                                  pa."service-hours",
+                                  pa."service-exceptions",
+                                  ((pa."pick-up-address").street,
+                                   (pa."pick-up-address").postal_code,
+                                   (pa."pick-up-address").post_office,
+                                   (pa."pick-up-address").postal_code
+                                  )::address,
+                                  pa."service-hours-info")::pick_up_location)
+             FROM "transport-service" service
+                      join lateral unnest((service.rentals)."pick-up-locations") pa on true
+    WHERE service.type = 'rentals' AND ts.id = service.id
+                         GROUP BY service.id);
 END;
 $$ LANGUAGE plpgsql;
 
--- a/database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
+++ b/database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
@@ -27,10 +27,11 @@ BEGIN
                                    (pa."pick-up-address").postal_code
                                   )::address,
                                   pa."service-hours-info")::pick_up_location)
-             FROM "transport-service" service
-                      join lateral unnest((service.rentals)."pick-up-locations") pa on true
-    WHERE service.type = 'rentals' AND ts.id = service.id
-                         GROUP BY service.id);
+               FROM "transport-service" service,
+                    lateral unnest((service.rentals)."pick-up-locations") pa
+              WHERE service.type = 'rentals'
+                AND ts.id = service.id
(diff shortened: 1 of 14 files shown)


database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
```diff
@@ -0,0 +1,65 @@
+ALTER TYPE address
+    ADD ATTRIBUTE "foreign_postal_code" VARCHAR(10);
+
+-- Create backup for "transport-service" table
+CREATE TABLE "transport-service_backup" AS TABLE "transport-service";
```
philosotractor: Was this table intended to be saved or should it be dropped in the PR?

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
```diff
@@ -0,0 +1,65 @@
...
+
+CREATE OR REPLACE FUNCTION convert_transport_service_address_postalcodes()
```
philosotractor: Here a sql function is created and stored to db scema.
Is that needed only once ever? If so, instead of creating and storing a function coudn't the same code be  run here using a `DO` block? Example in e.g. file `/mmtis-national-access-point/database/src/main/resources/db/migration/V1_135__detected_route_change_table.sql`

philosotractor: Alternatively, if `DO`is not used should the function be dropped in the end of the migration script?

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
```diff
@@ -0,0 +1,65 @@
...
+                    lateral unnest((service.rentals)."pick-up-locations") pa
+              WHERE service.type = 'rentals'
+                AND ts.id = service.id
+              GROUP BY service.id);
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT convert_transport_service_address_postalcodes();
+
+CREATE OR REPLACE FUNCTION convert_transport_operator_address_postalcodes()
```
philosotractor: Same comment as for previous function :)

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
```diff
@@ -0,0 +1,65 @@
...
+              FROM "transport-operator" top
+            WHERE top.id = "t-operator".id);
+
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT convert_transport_operator_address_postalcodes();
+
+ALTER TYPE address
+    DROP ATTRIBUTE "postal_code";
```
philosotractor: Doesn't dropping composite type's attributes cause trouble when there's already data in db created using that template?
E.g. in this file `rentals."pick-up-locations"` migration puts postal_code in the middle and foreign_postal_code in the end. Won't dropping postal_code lead to data in the db array not matching to the type anymore?

ote/src/cljc/ote/util/transport_operator_util.cljc
```diff
@@ -9,4 +9,9 @@
         filename (string/lower-case (str
                                       (string/replace operator-name " " "_")
                                       "_gtfs.zip"))]
-    filename))
\ No newline at end of file
+    filename))
+
+(def every-postal-code-regex #"(?i)^[a-z0-9][a-z0-9\- ]{0,9}$")
```
philosotractor: "_Sallitaan syötteeksi jatkossa numeroiden lisäksi esim. kirjaimia ja ainakin merkki "-". Minimimerkkimäärä 1 ja maksimimerkkimääräksi esim 16._"
1) regex  allows: 1-10 chars. Just making sure, is 10 intentional or should it be 16?
2) regex allows strings like `a---------`. 
Without exact rules that's as good as any, but would following suit better? 
```suggestion
(def every-postal-code-regex #"(?i)^(?=.{1,10}$)[a-z0-9]+-?[a-z0-9]+$")
```
Allows total string length of 1-10 for a string consisting of: 
one or more a-z0-9
zero or one dash
one or moreaz0-9

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
rckopteri: This is for backup purposes. I'm a bit afraid to update all address data.

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
rckopteri: I'll drop function afterwards

database/src/main/resources/db/migration/V1_169__change_address_post_code.sql
rckopteri: Dropping column from composite type seems to be ok. It works fine. Though I needed to update our test data sql files.

ote/src/cljc/ote/util/transport_operator_util.cljc
rckopteri: Does this allow empty spaces? NNN NNN ?

rckopteri: It didn't but I fixed it.

ote/src/cljc/ote/util/transport_operator_util.cljc
```diff
@@ -11,7 +11,7 @@
                                       "_gtfs.zip"))]
     filename))
 
-(def every-postal-code-regex #"(?i)^[a-z0-9][a-z0-9\- ]{0,9}$")
+(def every-postal-code-regex #"(?i)^(?=.{1,10}$)[a-z0-9][a-z0-9\- ]*[a-zA-Z0-9_]$")
```
philosotractor: Great! Just remove the extra _

PR #1193, opened 2019-10-11
168621293 select country fixes

rckopteri:
# Fixed
* Fixed transport-operator country-list flag check
* Remove country from address ( test database is cleaned up earlier. )


philosotractor:
LGTM, though didn't go through the state handling in my mind.

Which leads to the remark that 
country-list is not mandatory to be stored in appstate. 
An alternative approach is getting it via `get-country-list` each time where it's needed (a quite minor change)
The minor benefit would be not having to make sure country-list is initialized in appstate when it's needed.

rckopteri:
I removed countries from app-state

PR #1194, opened 2019-10-15
169115839 old routes have wrong data in calendar

rckopteri:
# Fixed
* Transit visualization view: Fixed calendar traffic by adding detection date as a limit in used package list
* Transit visualization view: Fixed url when route was selected in /all/ mode

### DIFF
--- a/database/src/main/resources/db/migration/R__GTFS_query.sql
+++ b/database/src/main/resources/db/migration/R__GTFS_query.sql
@@ -184,9 +184,7 @@ SELECT array_agg(x.id)
          ORDER BY "external-interface-description-id", created DESC) x
 $$ LANGUAGE SQL STABLE;
 
--- New sproc for getting packages is need when calendar wants to show what kind of traffic was back
--- in the day. We need to set detection_date to pick only those packages that were downloaded when detection
--- was run.
+-- Returns list of packages for service_id, which were the latest on detection_date. I.e. excludes those downloaded after detection_date.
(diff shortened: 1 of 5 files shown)


database/src/main/resources/db/migration/R__GTFS_query.sql
```diff
@@ -184,6 +184,22 @@ SELECT array_agg(x.id)
          ORDER BY "external-interface-description-id", created DESC) x
 $$ LANGUAGE SQL STABLE;
 
+-- New sproc for getting packages is need when calendar wants to show what kind of traffic was back
+-- in the day. We need to set detection_date to pick only those packages that were downloaded when detection
+-- was run.
```
philosotractor: Part of the comments might look a bit out-of-place after merging the PR. Following might be more suited here:

-- Returns list of packages for service_id, which were the latest on detection_date. I.e. excludes those downloaded after detection_date.

philosotractor:
Not sure if window location should be mangled via js after it's set wrong, 
instead existing router should be used so that url is set right in the first place. Will take a deeper look on that a bit later.

philosotractor:
The "/all" issue with the url seems to be fixed by appending "/all/" in
ote.views/transit_visualization.cljs/gtfs-package-info instead of "/all".

philosotractor:
The url "all" fix should be separate and not in this PR, because it doesn't seem trivial enough.

PR #1198, opened 2019-10-18
168477136 - No stops or trips in transit visualization page

rckopteri:
# Fixed
* Fixed situation where transit visualization page could not show stops or trips. Now we return first stop as first-common-stop if there is no common stops trips

### DIFF
--- a/ote/src/cljc/ote/transit_changes.cljc
+++ b/ote/src/cljc/ote/transit_changes.cljc
@@ -104,12 +104,21 @@
                           (count right-different-stop-keys))}))
 
 
-(defn time-for-stop [stoptimes-display stop-name]
-  (some #(when (= stop-name (:gtfs/stop-name %))
-           (:gtfs/departure-time %))
-        (:stoptimes stoptimes-display)))
+(defn time-for-stop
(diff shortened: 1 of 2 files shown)


ote/src/cljc/ote/transit_changes.cljc
```diff
@@ -104,12 +104,21 @@
                           (count right-different-stop-keys))}))
 
 
-(defn time-for-stop [stoptimes-display stop-name]
-  (some #(when (= stop-name (:gtfs/stop-name %))
-           (:gtfs/departure-time %))
-        (:stoptimes stoptimes-display)))
-
-(defn first-common-stop [stoptime-displays]
+(defn time-for-stop
```
philosotractor: Would following implementation "flow" better?
```
(defn time-for-stop
"..."
[stoptimes-display stop-name]
  (as-> (:stoptimes stoptimes-display) tmp
        (some #(when (= stop-name (:gtfs/stop-name %))
               (defn first-common-stop [stoptime-displays]
                 (:gtfs/departure-time %))
               tmp))
        (or tmp (:gtfs/departure-time (first (:stoptimes stoptimes-display))))))
```

PR #1204, opened 2019-10-24
169174702 netex tag to service search and refactoring service view netex link

philosotractor:
# Added
* netex: netex_util.clj for netex-related data mangling
* views: service_viewer: netex tag to those service interfaces having it

# Fixed
* Some problem fixed.
   
# Changed
* services: service-search: refactoring search handling
  *  service-search: append netex urls to service interfaces 
*  netex: interface id to netex filename allows multiple s3 files per service
* db: service_search_result_external_interface_id

+ Code cleanup here and there

### DIFF
--- a/database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
+++ b/database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
@@ -1,0 +1,49 @@
+ALTER TYPE external_interface_search_result
+    ADD ATTRIBUTE id INTEGER;
+
+DROP VIEW transport_service_search_result;
+
+-- Re-create view to include external-interface-description id to external_interface_search_result
+CREATE VIEW transport_service_search_result AS
+SELECT t.*,
+       op.name as "operator-name",
+       op."business-id" as "business-id",
+       -- Changed Start
+       array_cat(
+               array_cat(
+                       (SELECT sc."companies"
+                        FROM "service_company" sc
+                        WHERE sc."transport-service-id" = t.id),
+                       t."companies"),
+               array_agg((aso.name, aso."business-id")::company)) AS "service-companies",
+       -- Changed End
+       (SELECT array_agg(oaf."operation-area")
+        FROM "operation-area-facet" oaf
+        WHERE oaf."transport-service-id" = t.id) AS "operation-area-description",
+       (SELECT array_agg(
+                       ROW(ei."external-interface",
+                           ei.format,
+                           ei."ckan-resource-id",
+                           ei.license, ei."data-content",
+                           eids."download-error",
+                           eids."db-error",
+                           ei."id")::external_interface_search_result)
+        FROM "external-interface-description" ei
+                 LEFT JOIN (SELECT DISTINCT ON (eids."external-interface-description-id") eids."external-interface-description-id",
+                                                                                          eids.id,
+                                                                                          eids."db-error",
+                                                                                          eids."download-error"
+                            FROM "external-interface-download-status" eids
+                            ORDER BY eids."external-interface-description-id" DESC,
+                                     eids.id DESC)
+            eids ON ei.id = eids."external-interface-description-id"
+        WHERE ei."transport-service-id" = t.id)::external_interface_search_result[] AS "external-interface-links"
+FROM "transport-service" t
+         LEFT JOIN (SELECT top.name as name,
+                           top."business-id" as "business-id",
+                           a."service-id"
+                    FROM "transport-operator" top,
+                         "associated-service-operators" a
+                    WHERE a."operator-id" = top.id) as aso ON aso."service-id" = t.id
+         JOIN  "transport-operator" op ON op.id = t."transport-operator-id"
+GROUP BY t.id, op.name, op."business-id";
--- a/database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
+++ b/database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
@@ -8,7 +8,6 @@ CREATE VIEW transport_service_search_result AS
 SELECT t.*,
        op.name as "operator-name",
        op."business-id" as "business-id",
-       -- Changed Start
        array_cat(
                array_cat(
                        (SELECT sc."companies"
@@ -16,7 +15,7 @@ SELECT t.*,
                         WHERE sc."transport-service-id" = t.id),
                        t."companies"),
                array_agg((aso.name, aso."business-id")::company)) AS "service-companies",
-       -- Changed End
+       -- Changed code Starts
        (SELECT array_agg(oaf."operation-area")
         FROM "operation-area-facet" oaf
         WHERE oaf."transport-service-id" = t.id) AS "operation-area-description",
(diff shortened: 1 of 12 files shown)


ote/src/cljs/ote/views/service_search.cljs
```diff
@@ -97,12 +95,16 @@
...
        (map-indexed
-         (fn [i {::t-service/keys [external-interface format data-content] :as row}]
+         (fn [i {::t-service/keys [external-interface format data-content]
+                 :keys [url-ote-netex] :as row}]
            (let [data-content (if (nil? data-content)
                                 (::t-service/url external-interface)
                                 (parse-content-value data-content))]
              [:div {:key (str i "-" id)}
-              [common-ui/information-row-default data-content [:span (str/join format) [gtfs-viewer-link row]]]]))
+              [common-ui/information-row-default data-content [:span (str/join format)
```
rckopteri: Should than when wrap to str?
Like: (str (str/join format) (when url-ote-netex ", NeTEx"))

ote/src/clj/ote/services/transport.clj
```diff
@@ -346,7 +333,10 @@
       (http/no-cache-transit-response
         (-> (assoc ts ::t-service/operation-area
                       (places/fetch-transport-service-operation-area db id))
-            (append-ote-netex-interfaces config db)))
+
+            vector
+            (netex-util/append-ote-netex-urls config db ::t-service/external-interfaces)
+            first))
```
rckopteri:
```suggestion
            (as-> (assoc ts ::t-service/operation-area
                                  (places/fetch-transport-service-operation-area db id)) ts	                     	
                       (first (netex-util/append-ote-netex-urls (vector ts) config db ::t-service/external-interfaces)))
```

ote/src/clj/ote/services/transport.clj
```diff
@@ -103,7 +92,9 @@
       (http/no-cache-transit-response
         (-> (assoc ts ::t-service/operation-area
                       (places/fetch-transport-service-operation-area db id))
-            (append-ote-netex-interfaces config db))
+            vector
+            (netex-util/append-ote-netex-urls config db ::t-service/external-interfaces)
+            first)
         200)
```
rckopteri:
```suggestion
(http/no-cache-transit-response
        (as-> (assoc ts ::t-service/operation-area	        
                      (places/fetch-transport-service-operation-area db id)) ts	                      	           
            (first (netex-util/append-ote-netex-urls (vector ts) config db ::t-service/external-interfaces)))
            200)
```

ote/src/clj/ote/netex/netex_util.clj
```diff
@@ -0,0 +1,80 @@
+(ns ote.netex.netex_util
+  "Utilities for working with netex objects"
+  (:require [specql.core :as specql]
+            [specql.op :as op]
+            [ote.db.netex :as netex]
+            [ote.db.transport-service :as t-service]
+            ))
```
rckopteri: Indention

ote/src/clj/ote/netex/netex_util.clj
```diff
@@ -0,0 +1,80 @@
...
+  `config` = object which contains ote configuration properties for the environment
+  `db`= ote database
+  `ext-ifs-key` = A key, within a service object, which contains external interface objects collection
+  Return: collection of services with an ote-generated netex appended to those external interfaces for which a netex url is available."
+  [services config db ext-ifs-key]
+  (let [conversions (fetch-conversions-for-services config
+                                                    db
+                                                    (set (map #(::t-service/id %)
+                                                              services)))]
+    (reduce
```
rckopteri: Just wondering if this could have been made with a simple for loop?

database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
```diff
@@ -0,0 +1,49 @@
...
+    ADD ATTRIBUTE id INTEGER;
+
+DROP VIEW transport_service_search_result;
+
+-- Re-create view to include external-interface-description id to external_interface_search_result
+CREATE VIEW transport_service_search_result AS
+SELECT t.*,
+       op.name as "operator-name",
+       op."business-id" as "business-id",
+       -- Changed Start
```
rckopteri: move these comments to row 23 and end to 31

ote/src/clj/ote/netex/netex_util.clj
philosotractor: Ok

database/src/main/resources/db/migration/V1_174__service_search_result_external_interface_id.sql
philosotractor: Ok

ote/src/clj/ote/services/transport.clj
philosotractor: Thanks! 
However could you elaborate the reasoning for this suggestion? 
I guess this boils down to personal preference, initial motivation for current implementation was to have less nesting which usually simplifies reading

ote/src/cljs/ote/views/service_search.cljs
philosotractor: Ok

ote/src/clj/ote/services/transport.clj
rckopteri: Less nesting is always better but the single row "first" call is ugly for my eyes :)

ote/src/clj/ote/netex/netex_util.clj
philosotractor: Simple is always the best :)
What kind of loop did you have in mind?
Approach here was that `services` is not iterated because there might be a lot of them, and in any case typically there's less `conversions` than `services`. It seemed more optimal to go through the `conversions` and find the service matching to it.

From readability point of view iterating services in a `mapv` and in one such iteration iterating all `conversions` might be better. There's just the matter of possibility of several conversions per service, which has to be handled there as well.
I can try re-implementing this by having the main iteration to go through the `services`, if that was what you had in mind.

rckopteri: I'm used to think that reduce is needed when the original map must be changed. e.g. split services to multiple different objects.
But in this case we need only to add item/items to external-interfaces vector that is already in the original map (services / single service)

philosotractor: `reduce` passes state information from one iteration to another, the nature of state is up to the implementor. Transforming maps to other kind of maps is best done using map(v) function or `for`.
But that's besides the point, 
the more important design decision is which collection to iterate. 
This probably won't look any simpler if reduce is replaced using a mapv, the main problem will boil down to the nested iteration. I can do an alternative, but it might be you won't like it any better.

philosotractor: > But in this case we need only to add item/items to external-interfaces vector that is already in the original map (services / single service)
That's not the case, here iteration should add a _key_ under some of the objects in the external-interfaces vector. No new items i.e. objects are added to the vector.

philosotractor: If you don't mind, I'll merge this and let's check the improved loop together to avoid using time to something you don't want it to be used :)

rckopteri: 👍

PR #1208, opened 2019-10-25
admin view table content wrapping and minor styling tweaks

philosotractor:
# Changed
* admin view table content wrapping and minor styling tweaks

### DIFF
--- a/ote/src/cljs/ote/views/admin.cljs
+++ b/ote/src/cljs/ote/views/admin.cljs
@@ -6,10 +6,9 @@
             [cljs-react-material-ui.icons :as ic]
             [reagent.core :as r]
             [stylefy.core :as stylefy]
-
+            [ote.ui.buttons :as buttons]
             [ote.db.transport-operator :as t-operator]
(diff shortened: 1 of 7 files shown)


ote/src/cljs/ote/views/admin.cljs
```diff
@@ -78,60 +77,94 @@
...
+                       {:width "18%"})
+                email]
+               [ui/table-row-column
+                (merge (stylefy/use-style style-base/table-col-style-wrap)
+                       {:width "20%"})
+                [ui/icon-button {:href "#"
+                                 :on-click #(do
+                                              (.preventDefault %)
+                                              (e! (fp/->ChangePage :operator-users {:ckan-group-id ckan-group-id})))}
+                 [ic/social-people]]
```
rckopteri: Is this change something that Jussi suggested?

philosotractor: No, but new button describes the actions better. Moved under actions title along other action buttons, because it's an action button :)

PR #1210, opened 2019-10-28
168662847 stat service - new db table

rckopteri:
# Added
* netex download: Added download statistics to database
* chouette.zip to tools/chouette
   
# Fixed
* service-id vs. interface-id confusion in netex download url generation
   
# Datamodel
* New enum stats-type
* New db table stats-service


tools/chouette/initialize-chouette.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+echo "1. Create /opt/chouette/ folder"
+if [ ! -d "/opt/chouette/" ]; then
+
+  sudo mkdir /opt/chouette
```
philosotractor:
```suggestion
  sudo mkdir -p /opt/chouette
```
-p switch creates whole path structure, but also as a side effect does not complete using an error if directory already exists.

tools/chouette/initialize-chouette.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "Copy - FAIL"
+fi
+
+echo "3. Unzip chouette.zip"
+sudo cd /opt/chouette
+sudo unzip /opt/chouette/mobi.chouette.command-3.4.8.zip -d /opt/chouette/
+
+echo "4. Give permissions "
+sudo cd ..
+sudo chown -R markusva:staff /opt/chouette
```
philosotractor: Something alse than `markusva` here... :)
```suggestion
sudo chown -R markusva:staff /opt/chouette
```

If user can't be resolved then would `chmod ugo+rwx` be ok...?

tools/chouette/initialize-chouette.sh
```diff
@@ -0,0 +1,41 @@
...
+sudo cp mobi.chouette.command-3.4.8.zip /opt/chouette/
+if [ $? -eq 0 ]; then
+    echo "Copy - OK"
+else
+    echo "Copy - FAIL"
+fi
+
+echo "3. Unzip chouette.zip"
+sudo cd /opt/chouette
+sudo unzip /opt/chouette/mobi.chouette.command-3.4.8.zip -d /opt/chouette/
```
philosotractor: zip name/path could be given as arg to script to make it universal across versios

tools/chouette/initialize-chouette.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "4. Give permissions "
+sudo cd ..
+sudo chown -R markusva:staff /opt/chouette
+sudo chmod -R 777 /opt/chouette
+
+echo "5. Remove .zip"
+rm -rf /opt/chouette/mobi.chouette.command-3.4.8.zip
+
+echo "6. Copy files to /opt/chouette"
+sudo mv /opt/chouette/chouette-cmd_3.4.8/* /opt/chouette/
```
philosotractor: While this is of course completely ok, an alternative approach is taken in
/mmtis-national-access-point/.circleci/Dockerfile
where zip is extracted to /tmp/chouette/ and content moved from /tmp/chouette/chouette-cmd_3.4.8 to /tmp/chouette. The difference is that all file system operations there happen within the chouette folder and there's no possibility to have orphan sibling dorectories if something goes wrong in the script.

ote/src/cljc/ote/db/common.cljc
```diff
@@ -10,5 +10,12 @@
...
 (define-tables
+  ;; Enums
+  ["stat_type" ::stat_type (specql.transform/transform (specql.transform/to-keyword))]
+
   ;; Common UDT types
-  ["address" ::address])
+  ["address" ::address]
+
+  ;; Tables
+  ["stats-service" ::stats-service]
```
philosotractor: If stats-service is a generic table, shouldn't it and its types be defined in a new `ote.db.stats`?
(or if it's netex-specific it could go into `ote.db.netex`)

ote/src/clj/ote/netex/netex_util.clj
```diff
@@ -39,25 +39,23 @@
...
-                                                                         (::netex/transport-service-id conversion))
-                                                                  conversion))
-                                                              conversions)]
-                            (assoc interface :url-ote-netex
-                                             (file-download-url config
-                                                                service-id
-                                                                (::netex/id interface-conversion)))
-                            interface))))))
-          services)))
+  (let [conversions (fetch-conversions-for-services config db (set (map #(::t-service/id %)
```
philosotractor: Indentation for function body should be two spaces. :)

ote/src/clj/ote/netex/netex_util.clj
```diff
@@ -39,25 +39,23 @@
...
+        (mapv (fn [service]
+                (update service
+                        ext-ifs-key
+                        (fn [interfaces]
+                          (vec
+                            (for [interface interfaces
+                                  :let [interface-id (::t-service/id interface)]]
+                              (if-let [interface-conversion (some (fn [conversion]
+                                                                    (when (= interface-id
+                                                                             (::netex/external-interface-description-id conversion))
```
philosotractor: 👍

ote/src/cljc/ote/db/common.cljc
rckopteri: Maybe we should define new ote.db.stats because it will be used in many cases. It is more related to service (stats-service) than netex and that is why I would not put it in ote.db.netex.

I thought to add it to ote.db.transport-service but I think that there is just too much data already. 

If ote.db.common is the wrong place I can move it.

ote/src/clj/ote/integration/export/netex.clj
```diff
@@ -36,10 +51,13 @@
         (http/publish! http
                        {:authenticated? false}
                        (GET "/export/netex/:transport-service-id{[0-9]+}/:file-id{[0-9]+}"
-                            {{:keys [file-id]} :params}
+                            {{:keys [file-id transport-service-id]} :params
+                             {:strs [origin]} :query-params}
```
philosotractor: 👍

ote/src/clj/ote/integration/export/netex.clj
```diff
@@ -7,24 +7,39 @@
...
+            [specql.core :as specql]
+            [ote.db.common :as common]))
 
 (defn file-download-url [{{base-url :base-url} :environment} transport-service-id file-id]
   (format "%sexport/netex/%d/%d" base-url transport-service-id file-id))
 
-(defn fetch-netex-response [db {:keys [bucket]} file-id]
+(defn fetch-netex-response
+  "Return netex file as reponse.
+  `origin` parameter is not mandatody. If it is missing, it is handled as api call."
```
philosotractor:
```suggestion
  `origin` parameter is not mandatory. If it is missing, it is handled as api call."
```

ote/src/clj/ote/integration/export/netex.clj
```diff
@@ -7,24 +7,39 @@
...
+               (s3/get-object bucket filename))
+        stats-type (if (or (nil? origin) (= "api" origin))
+                     :download-route-netex-api
+                     :download-route-netex-ui)]
+    (do
+      ;; Store download statistics to database
+      (specql/insert! db ::common/stats-service
+                      {::common/transport-service-id service-id
+                       ::common/type stats-type
+                       ::common/created (java.sql.Timestamp. (System/currentTimeMillis))})
```
philosotractor: Which should be used, this or `java.util.Date.` ? Just asking.

ote/src/cljc/ote/db/common.cljc
philosotractor: Agreed, only pro's when putting in ote.db.stats

tools/chouette/initialize-chouette.sh
rckopteri: haha. thanx!
Fixed by getting current user instead of hardcoded value.

PR #1211, opened 2019-10-28
169393181 download table is missing an index

rckopteri:
# Added
* db index to external-interface-download-status table
* Add check to netex file download url generation to not to generate url if there is no netex conversions.


ote/src/clj/ote/netex/netex_util.clj
```diff
@@ -39,25 +39,26 @@
...
-                            (assoc interface :url-ote-netex
-                                             (file-download-url config
-                                                                service-id
-                                                                (::netex/id interface-conversion)))
-                            interface))))))
-          services)))
+  (let [conversions
+        (fetch-conversions-for-services config db (set (map #(::t-service/id %)
+                                                            services)))]
+    (if (some? conversions)
```
philosotractor: 👍

PR #1216, opened 2019-10-30
169287614 dev netex documentation

rckopteri:
# Added
* Added new netex/chouette help file
   
# Fixed
* Search service api documentation
* log/info to log/error when there is an actual error in laundry service
   
# Changed
* Small update to main readme file


docs/api/README.md
```diff
@@ -61,17 +60,27 @@ The following shows an example of the response in the JSON format:
...
       "sub-type": "terminal",
-      "id": 2
+      "id": 2,
+      "published":"1970-01-01T00:00:00Z",
+      "description":[
+        {"lang":"FI","text":"Bussipysäkki keskellä suomea"}, 
+        {"lang":"EN","text":"Bus stop in the middle of nowhere"}],
+      "business-id":"1234567-0",
+      "external-interface-links":[
+      {"external-interface":
```
philosotractor: :url-ote-netex is not mentioned here
but in the other hand it's not in schema either:
https://testi.finap.fi/export/geojson/transport-service.schema.json

But it _is_ in geoJson export, e.g. https://testi.finap.fi/export/geojson/497/1691
I guess we can overlook this for now if you like.

docs/devsetup-netex.md
```diff
@@ -0,0 +1,11 @@
+### How to setup netex conversion in development environment
```
philosotractor: Ok, but if you like we could mention the netex unit test as well in section 3

docs/api/README.md
rckopteri: 👍

docs/devsetup-netex.md
rckopteri: 👍

PR #1325, opened 2021-11-10
New developer modifications

esuomi:
# Fixed
* Use container internal service name for testdata database host instead of MacOS specific localhost resolver
* fix repository references for database migration scripts
   
# Changed
* remove duplicate clj-kondo/ entry


rckopteri (approved):
👍

PR #1326, opened 2021-11-18
Finap 14/colors fonts layout

antti-nuortimo-solita:
# Changed
* Changed topnav background color to black.
* Changed topnav dropdown background color to black.
* Changed footer background color to black.
* Changed primary font to Public Sans.

### DIFF
--- a/ote/resources/public/css/styles.css
+++ b/ote/resources/public/css/styles.css
@@ -238,7 +238,7 @@ footer {
 
 }
 .site-footer {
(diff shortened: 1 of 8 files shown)


esuomi:
Seems this will also impact sent email layouts as well `ote/src/clj/ote/util/email_template.clj` but I think that's actually a good thing :) lgtm 👍

rckopteri (approved):

PR #1332, opened 2021-11-24
Finap 14/colors

antti-nuortimo-solita:
# Added
* Parametrized brand colors.
   
# Changed
* Use parametrized brand colors on front page in header.
* Use parametrized brand colors on front page in buttons.

### DIFF
--- a/ote/src/cljs/ote/theme/colors.cljs
+++ b/ote/src/cljs/ote/theme/colors.cljs
@@ -14,6 +14,7 @@
 
 (def primary-background-color basic-black)
 (def primary-text-color basic-white)
+(def primary-button-background-color dark-blue)
 
 (def blue-lighter "#66a3e0")
 (def blue-light "#3385d6")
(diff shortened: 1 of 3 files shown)


ote/src/cljs/ote/theme/colors.cljs
```diff
@@ -1,5 +1,21 @@
 (ns ote.theme.colors)
 
+(def dark-blue "#0034ac")
+(def dark-green "#005f61")
+(def basic-purple "#73468c")
+(def dark-purple "#520076")
+
+(def dark-red "#b40000")
+(def dark-gray "#505064")
+(def basic-brown "#784b28")
```
esuomi, line 10: This is just a note, no changes needed - once the whole palette is here, I'd reorder these into some logical order, eg. all greens in order basic-dark-light, or first basic colors, then dark colors, then light colors. Now they're in a bit of mix&match order.

esuomi (approved):
Just one small thing to note for future, looks good so far 👍

ote/src/cljs/ote/theme/colors.cljs
antti-nuortimo-solita: Currently they are in order given in the brand book, then black and white, which actually could be first before these accessible colors. But good idea and I'll make the order more logical in the near future.

PR #1334, opened 2021-11-25
inline feather icons dependency

esuomi:
inline tatut/re-svg-icons:feather-icons to project to avoid NoClassDefFoundError in build process

Leiningen plugins cannot be set to ignore transitive dependencies and since re-svg-icons isn't a publically available JAR, the only option left is to include the relevant code directly into the project.
   
# Fixed
* feather icon dependency causing build errors in CI


antti-nuortimo-solita (approved):
I haven't tested this but looks good to me :ship:

PR #1338, opened 2021-11-26
add collapsible two-level menu for mobile layout

esuomi:
# Added
* collapsible two-level menu for mobile layout

---

Screenshot in three parts:
 1. What it was like before (logged in)
 2. What it is now when collapsed (logged out)
 3. What it is now when expanded fully (logged out)

image


antti-nuortimo-solita (approved):
Not tested, but looks good to me :ship:

PR #1339, opened 2021-11-26
FINAP-12/final touches

esuomi:
bits and pieces from eg. demo and such


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1341, opened 2021-11-29
Use primary button type for "Add new report of change" button.

antti-nuortimo-solita:
# Changed
* Use primary button type for "Add new report of change" button.


esuomi (approved):
We had a quick telco about this, looks nice now 👍

PR #1342, opened 2021-11-29
Fix error in CSS which caused page to render incorrectly

esuomi:
# Fixed
* Footer CSS was broken by tidy-up in another branch


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1344, opened 2021-11-30
Replace all nap@traficom.fi references with nap@fintraffic.fi

esuomi:
This is just a simple search&replace, translation updates and eg. phone numbers will follow separately.

# Changed
* Replace all nap@traficom.fi references with nap@fintraffic.fi


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1346, opened 2021-12-01
Use button text in Finnish as is in Swedish and English translations.

antti-nuortimo-solita:
# Changed
* Use button text "Lataa liikkumispalveluiden tarjoajien tiedot CSV:nä" in Monitor view in Swedish and English as well.


esuomi (approved):
For reference, it was discussed that there's no need to translate these so this is fine 👍

PR #1348, opened 2021-12-02
FINAP-5/remove localstorage tos

esuomi:
# Changed
* old blue topbar TOS is entirely removed, services search has an additional TOS notification text

 - [x] Official Swedish and English translations still missing

### DIFF
--- a/ote/resources/public/language/sv.edn
+++ b/ote/resources/public/language/sv.edn
@@ -1254,7 +1254,8 @@ Tjänsterna beskrivs var för sig på egen blankett. I exemplet fyller man i bå
   :service-count [:count " tjänster"]
   :service-providing-text ", i vars produktion medverkar totalt "
   :service-and-operator-summary-text [:count " företag eller organisationer."]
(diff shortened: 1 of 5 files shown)


ote/resources/public/language/sv.edn
```diff
@@ -1254,6 +1254,8 @@ Tjänsterna beskrivs var för sig på egen blankett. I exemplet fyller man i bå
   :service-count [:count " tjänster"]
   :service-providing-text ", i vars produktion medverkar totalt "
   :service-and-operator-summary-text [:count " företag eller organisationer."]
+  :service-search-tos-notification-start "Hakemalla ja hyödyntämällä aineistoja sitoudut palvelun käyttöehtoihin."
```
antti-nuortimo-solita, line 1257: I assume a translation in Swedish would be needed here, from the customer.

antti-nuortimo-solita (commented):
Unless customer has confirmed that no Swedish translation is needed for :service-search-tos-notification-start, a Swedish translation would be needed from the customer.

Otherwise, looks good to me :ship:

PR #1350, opened 2021-12-07
FINAP-22/common link interceptor component for all links

esuomi:
# Changed
* even more generalized `linkify` component which detects external links, renders an _(optional)_ external link icon next to the link and adds an on-click nagger if the URL is not a known good URL

---

Screenshot 2021-12-07 at 15 43 31
Screenshot 2021-12-07 at 15 43 49

# TODO

 - [ ] translations, as always

### DIFF
--- a/ote/src/cljs/ote/ui/common.cljs
+++ b/ote/src/cljs/ote/ui/common.cljs
@@ -12,6 +12,7 @@
             [reagent.core :as r]
             [clojure.string :as str]
             [reagent.core :as reagent]
+            [re-svg-icons.feather-icons :as feather-icons]
 
             [goog.crypt.Md5]
             [goog.crypt]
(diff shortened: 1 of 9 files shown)


ote/src/cljs/ote/ui/common.cljs
```diff
@@ -12,58 +12,111 @@
...
+   ```"
+  [url]
+  (cond
+    ;; URL with protocol, use as is
+    (re-matches #"^(\w+:).*" url)
+    [url true]
+
+    ;; User specified link without protocol (like "www.serviceprovider.fi/foo")
+    (re-matches #"^[^/]+\..*" url)
+    [(str "http://" url) true]
```
antti-nuortimo-solita, line 69: Should `https://` be used here?

antti-nuortimo-solita (commented):
Except missing translations looks good to me :ship: 

There was one URL string "http://", I was wondering if "https://" is needed instead or in addition to it.

ote/src/cljs/ote/ui/common.cljs
esuomi: Not all websites (still! 😢) provide HTTPS, so that would generate more broken links than the current option - most sites do autoredirect from http to https, thankfully. Since these are just strings we can't really do much about it...

antti-nuortimo-solita: OK. Can we accept both "https://" and "http://"?

antti-nuortimo-solita: On the other hand, that's beyond the scope of this task of warning the user of external links.

PR #1352, opened 2021-12-09
FINAP-9/add Traficom infobox

esuomi:
# Added
* Traficom infobox to front page

---
Screenshot 2021-12-09 at 16 21 33


antti-nuortimo-solita (approved):
Looks good me, translations will follow later :ship:

PR #1356, opened 2021-12-30
FINAP-8/EU flag and funding notification

esuomi:
# Added
* EU flag and funding notification

---
Screenshot 2021-12-30 at 13 14 10


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1359, opened 2022-01-04
FINAP-60/more generic labels to avoid issue where not all taxi drivers have exactly the same schedules

esuomi:
Division between prices is now daytime, nighttime and weekends.

Screenshot 2022-01-04 at 12 12 19


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1361, opened 2022-01-04
FINAP-64/add simple warning header to the page to differentiate test environments from production

esuomi:
Visible in localhost and test.

Screenshot 2022-01-04 at 13 06 40


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1365, opened 2022-01-13
FINAP-61/bind statistics table to data returned by the backend

esuomi:
All statistics keys are now harmonized between various components ranging from translations to database column names.

Known issues:
 - Table cannot sorted by service provider name, since that's actually a concatenation of two joined columns and outside the evaluation scope of actual `ORDER BY` clause.
 - Area code column is missing
 - Example price column is missing

All these are documented in JIRA.

Screenshot 2022-01-13 at 15 43 25

### DIFF
--- a/ote/src/clj/ote/services/taxiui_service.sql
+++ b/ote/src/clj/ote/services/taxiui_service.sql
@@ -24,3 +24,17 @@ SELECT id,
 
 -- name: fetch-services-with-prices
 SELECT DISTINCT service_id FROM taxi_service_prices;
+
+-- name: list-pricing-statistics
+SELECT o.id AS "operator-id",
+       s.id AS "service-id",
+       CONCAT(o.name, '/', s.name) AS name,
+       l.timestamp AS timestamp,
+       l.start_price_daytime AS "start-price-daytime",
+       l.start_price_nighttime AS "start-price-nighttime",
+       l.start_price_weekend AS "start-price-weekend",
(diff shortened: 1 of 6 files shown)


ote/src/clj/ote/services/taxiui_service.sql
```diff
@@ -23,4 +23,18 @@ SELECT id,
...
+       CONCAT(o.name, '/', s.name) AS name,
+       l.timestamp AS timestamp,
+       l.start_price_daytime AS "start-price-daytime",
+       l.start_price_nighttime AS "start-price-nighttime",
+       l.start_price_weekend AS "start-price-weekend",
+       l.price_per_minute AS "price-per-minute",
+       l.price_per_kilometer AS "price-per-kilometer"
+  FROM list_taxi_pricing_statistics(:column, :direction) l
+  JOIN "transport-service" s ON l.service_id = s.id
+  JOIN "transport-operator" o ON s."transport-operator-id" = o.id;
```
antti-nuortimo-solita, line 40: For consistency it could be better to use quotes in all column names.

antti-nuortimo-solita (approved):
Looks good to me :ship:

ote/src/clj/ote/services/taxiui_service.sql
esuomi: The autoformatter seemed to quote only ambiguous columns, so yeah, I'll fix that.

PR #1366, opened 2022-01-14
FINAP-45/Add constraints for columns used for input data

antti-nuortimo-solita:
# Added
* Add constraints to database columns used to store input data.
* Constraints have been defined only for columns that are not empty, that is, they contain some data.


esuomi (approved):
Looks alright to me, of course validating each constraint is a task in itself. We'll have to be vigilant if any of these cause eg. customer feedback, for now it's 👍

PR #1368, opened 2022-01-17
FINAP-61/add primary operating areas to statistic view results

esuomi:
Adds primary operating areas to the table, and a two-level sorting to enable sort for both name and operating areas column.
Screenshot 2022-01-17 at 16 40 26

### DIFF
--- a/database/src/main/resources/db/migration/R__list_taxi_pricing_statistics.sql
+++ b/database/src/main/resources/db/migration/R__list_taxi_pricing_statistics.sql
@@ -1,4 +1,7 @@
--- Function for listing taxi service pricing statistics. Function wrapping is used to avoid massive, repeating
+-- Functions for listing taxi service pricing statistics.
+-- list_taxi_statistics(...) is meant to be called directly, other functions are helpers
+
+-- Function wrapping is used to avoid a massive, repeating
 -- ORDER BY clause; the quote_ident etc. keeps sure this function is safe from SQLi.
 --
 -- Original from https://stackoverflow.com/a/8146245/44523, modified to our needs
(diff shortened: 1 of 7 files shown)


database/src/main/resources/db/migration/R__list_taxi_pricing_statistics.sql
```diff
@@ -12,8 +15,73 @@ BEGIN
...
+        SELECT o.id AS "operator-id",
+               s.id AS "service-id",
+               CONCAT(o."name", ''/'', s."name") AS "name",
+               l.timestamp AS "timestamp",
+               l.start_price_daytime AS "start-price-daytime",
+               l.start_price_nighttime AS "start-price-nighttime",
+               l.start_price_weekend AS "start-price-weekend",
+               l.price_per_minute AS "price-per-minute",
+               l.price_per_kilometer AS "price-per-kilometer",
+               (SELECT array_agg(oa_d.text)
```
antti-nuortimo-solita, line 59: My IDEA formats `array_agg` as `ARRAY_AGG`... though PostgreSQL's own documentation uses `array_agg`.

esuomi: I suppose it wants to capitalize built-in functions to contrast them from user made functions. I'd rather not change this at this point, I need to take a step back from this horrid concatenation to get a bit better perspective on if it's even good... :)

antti-nuortimo-solita (commented):
Quite complicated SQL code. Not tested. Unit tests would be good. Otherwise, looks good to me :ship:

PR #1370, opened 2022-01-18
FINAP-61/example price column

esuomi:
Example price column for each service.

Screenshot 2022-01-18 at 15 14 18


antti-nuortimo-solita (approved):
Maybe it would be good to tell the user how is the example trip (Start at daytime, 10 km, 15 min)?

Looks good to me :ship:

esuomi:
Yeah, the explanation should be visible somewhere, but the parts which were meant to show it have been cut from scope, so now it's just "magical"...lets see if we can figure out something else once I get to filters :)

PR #1371, opened 2022-01-18
FINAP-60/move tuck fx to load refreshed data only after save is successful to avoid desyncing UI

esuomi:
As it says on the title, noticed this while fiddling around with statistics view.


antti-nuortimo-solita (approved):
We discussed this on a call. Looks good to me :ship:

PR #1372, opened 2022-01-19
FINAP-62/price update age filter

esuomi:
Under six months, within a year, over a year ago filter for filtering ages.

The SQL horror expands, now there's query filters on three levels 😈 

Screenshot 2022-01-19 at 14 34 41

### DIFF
--- a/ote/src/clj/ote/services/taxiui_service.clj
+++ b/ote/src/clj/ote/services/taxiui_service.clj
@@ -90,17 +90,27 @@
               (some->> (:unknown-places places)
                        (log/warnf "Detected unknown places while updating %s/%s details! %s" operator-id service-id)))))))))
 
+(defn- age-filter
+  "Returns age filter as PostgreSQL INTERVAL compatible string."
+  [filters]
+  (case (:age-filter filters)
+    :within-six-months "6 months"
+    :within-one-year   "11 months 27 days"  ; this is a minor hack to allow easy "diff >= 1 year" selection in SQL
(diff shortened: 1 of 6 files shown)


ote/src/clj/ote/services/taxiui_service.clj
```diff
@@ -90,17 +90,27 @@
               (some->> (:unknown-places places)
                        (log/warnf "Detected unknown places while updating %s/%s details! %s" operator-id service-id)))))))))
 
+(defn- age-filter
+  "Returns age filter as PostgreSQL INTERVAL compatible string."
+  [filters]
+  (case (:age-filter filters)
+    :within-six-months "6 months"
+    :within-one-year   "11 months 27 days"  ; this is a minor hack to allow easy "diff >= 1 year" selection in SQL
```
antti-nuortimo-solita, line 98: Does this hack affect the functionality of the application (selection within 1 year)?

antti-nuortimo-solita (commented):
I'm wondering about the "minor hack", but assuming it's OK, then looks good to me :ship:

ote/src/clj/ote/services/taxiui_service.clj
esuomi: In a positive way, yes. The problem is expressing "less than 1 year" in same CASE in the SQL query as "more than 1 year" with a single boolean and a simple clause, and I didn't want to overcomplicate the query.

PR #1373, opened 2022-01-19
FINAP-62/add filtering results by company name to stats view

esuomi:
Also uses `goog.functions.debouncer` to prevent hammering the API character by character.

Screenshot 2022-01-19 at 16 50 02


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1375, opened 2022-01-19
FINAP-61/fix date range comparison by comparing only the filter interval

esuomi:
Previous logic, upon further inspection, didn't really make sense.


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1376, opened 2022-01-21
FINAP-61/Add params to function drop to disambiguate which function should be removed

esuomi:
Seems the function overloads turned ambiguous in test environment at least, so need to add params.


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1378, opened 2022-01-24
FINAP-62/create area-filter for reducing amount of results

esuomi:
This is quite a bit simpler than envisioned, and only matches one area at any given time. This is mostly a time saving optimization, and can be improved later on.


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1380, opened 2022-01-24
FINAP-64/make stats page require authentication

esuomi:
Quite a simple change :)


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1381, opened 2022-01-25
FINAP-64/front page data binding

esuomi:
Real data is always nicer than fake data.

Screenshot 2022-01-25 at 14 34 09

### DIFF
--- a/ote/src/clj/ote/services/taxiui_service.clj
+++ b/ote/src/clj/ote/services/taxiui_service.clj
@@ -127,13 +127,6 @@
 
 (defn fetch-service-summaries
   [db user {}]
-  [{:id "service-id-num"
-    :name "namenamename"
-    :updated "23.12.2021"
-    :example-trip 36.90
-    :price-per-kilometer 1.5
-    :price-per-minute 1
-    :operating-areas [{:label "Jokualue" :primary? true}]}]
   (let [groups (authorization/user-transport-operators db user)]
     (vec (->> (list-service-summaries db {:operator-ids groups})
               (map (fn [service] (update service :operating-areas #(db-util/PgArray->vec %))))))))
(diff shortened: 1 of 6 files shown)


ote/src/clj/ote/services/taxiui_service.clj
```diff
@@ -104,6 +104,7 @@
 
 (defn fetch-pricing-statistics
   [db {:keys [sorting filters]}]
+  ; TODO auth checks
```
antti-nuortimo-solita, line 107: Auth checks seem to be missing here.

antti-nuortimo-solita: And on line 125 as well.

ote/src/clj/ote/services/taxiui_service.clj
```diff
@@ -121,8 +122,22 @@
...
 
+(defn fetch-service-summaries
+  [db user {}]
+  [{:id "service-id-num"
+    :name "namenamename"
+    :updated "23.12.2021"
+    :example-trip 36.90
+    :price-per-kilometer 1.5
+    :price-per-minute 1
+    :operating-areas [{:label "Jokualue" :primary? true}]}]
```
antti-nuortimo-solita, line 136: These values above seem to be some testing data and would still need to be replaced with proper values, I guess.

esuomi: whoopsie, will remove

ote/src/clj/ote/services/taxiui_service.clj
esuomi: Funnily enough, the auth check is in the actual API endpoint definition below, just forgot to remove this TODO :) Too much coffee?

antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1382, opened 2022-01-26
Finap 95/session timeout

ghost:
# Changed
* Set `Expires` timestamp string for `auth_tkt` cookie.
* Upon logout set `Expires` timestamp string to Unix epoch.
* Use "UTC" as the zone id. The timestamp string follows the RFC 1123.
    Example: "Thu, 1 Jan 1970 00:00:00 GMT"
* Update java-time package.

Does not work properly yet ("Ongelma palvelinkutsussa"):
Screenshot from 2022-01-27 15-03-06

### DIFF
--- a/ote/src/clj/ote/services/login.clj
+++ b/ote/src/clj/ote/services/login.clj
@@ -28,7 +28,7 @@
 
 (defqueries "ote/services/login.sql")
 
-(defn- unix-epoch []
+(defn ^:private unix-epoch []
   ;; Returns a zoned date-time that can later be
   ;; formatted as: "Thu, 1 Jan 1970 00:00:00 GMT"
   (java-time.zone/zoned-date-time 1970 "UTC"))
--- a/ote/src/clj/ote/services/login.clj
+++ b/ote/src/clj/ote/services/login.clj
@@ -66,13 +66,10 @@
                              :user-id          (:id login-info)
                              :user-data        {:session-start-timestamp current-zoned-timestamp}})
             (:domain auth-tkt-config)
-            ;; Timestamp for cookie expiration.
-            (let
-              ; Define the session duration here.
-              [session-duration-in-hours 8]                 ; Define the session duration here!
-              ;; If you want to use minutes or seconds instead of hours,
-              ;; change the line before and after this to match the unit in use.
-              (.plus current-zoned-timestamp (java-time.amount/hours session-duration-in-hours)))
+            ; Define the absolute session timeout in hours.
+            ; If you want to use minutes or seconds instead of hours,
+            ; change the line below to match the unit in use.
+            (.plus current-zoned-timestamp (java-time.amount/hours 8))
             false))
 
         (http/transit-response {:error :unconfirmed-email} 401)) ;; This could be 403 instead
--- a/ote/src/clj/ote/services/login.clj
+++ b/ote/src/clj/ote/services/login.clj
@@ -29,7 +29,7 @@
 (defqueries "ote/services/login.sql")
 
 (defn- unix-epoch []
-  ;; Returns a zoned time date that can later be
+  ;; Returns a zoned date-time that can later be
   ;; formatted as: "Thu, 1 Jan 1970 00:00:00 GMT"
   (java-time.zone/zoned-date-time 1970 "UTC"))
 
--- a/ote/src/clj/ote/services/login.clj
+++ b/ote/src/clj/ote/services/login.clj
@@ -7,6 +7,7 @@
             [compojure.core :refer [routes POST wrap-routes]]
             [ote.nap.cookie :as cookie]
             [ote.nap.users :as users]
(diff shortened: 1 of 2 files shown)


ote/src/clj/ote/services/login.clj
```diff
@@ -27,16 +28,22 @@
...
-  (update response :headers
-          assoc "Set-Cookie" (if (nil? domain)
-                               [(str "auth_tkt=" auth-tkt-value "; Path=/; HttpOnly")]
-                               ;; Three cookies are required to match ckan cookie configuration
-                               [(str "auth_tkt=" auth-tkt-value "; Path=/; HttpOnly; Secure")
-                                (str "auth_tkt=" auth-tkt-value "; Path=/; HttpOnly"
-                                     "; Domain=" domain "; Secure")
-                                (str "auth_tkt=" auth-tkt-value "; Path=/; HttpOnly"
-                                     "; Domain=." domain "; Secure")])))
+(defn- unix-epoch []
```
esuomi, line 31: Could be just `(def ^:private unix-epoch ...)` since ability to call this doesn't produce alternate values over time.

ote/src/clj/ote/services/login.clj
```diff
@@ -45,27 +52,36 @@
...
+                 (transport-operator/get-user-transport-operators-with-services db (:groups user) (:user user)))}
+              200)
+            (cookie/unparse "0.0.0.0" (:shared-secret auth-tkt-config)
+                            {:digest-algorithm (:digest-algorithm auth-tkt-config)
+                             :timestamp        (java.util.Date.)
+                             :user-id          (:id login-info)
+                             :user-data        {:session-start-timestamp current-zoned-timestamp}})
+            (:domain auth-tkt-config)
+            ;; Timestamp for cookie expiration.
+            (let
```
esuomi, line 70: IMO this is too much wrapping, just have the `8` directly in the `.plus` call.

ote/src/clj/ote/services/login.clj
antti-nuortimo-solita: Fixed :)

ote/src/clj/ote/services/login.clj
antti-nuortimo-solita: Fixed :)

ote/src/clj/ote/services/login.clj
```diff
@@ -45,27 +52,33 @@
...
+            false))
 
         (http/transit-response {:error :unconfirmed-email} 401)) ;; This could be 403 instead
       ;; Login shall indicate error in a general way and not reveal which credential was invalid
       (http/transit-response {:error :login-error} 400))
     (http/transit-response {:error :login-error} 400)))
 
 (defn- logout [auth-tkt-config]
-  (with-auth-tkt (http/transit-response :ok) "" (:domain auth-tkt-config)))
+  (with-auth-tkt (http/transit-response :ok) "" (:domain auth-tkt-config) (unix-epoch) true))
```
esuomi, line 81: Remove parenthesis around `unix-epoch` since it's a value now :)

ote/src/clj/ote/services/login.clj
```diff
@@ -27,16 +28,22 @@
...
-                                     "; Domain=" domain "; Secure")
-                                (str "auth_tkt=" auth-tkt-value "; Path=/; HttpOnly"
-                                     "; Domain=." domain "; Secure")])))
+(defn ^:private unix-epoch []
+  ;; Returns a zoned date-time that can later be
+  ;; formatted as: "Thu, 1 Jan 1970 00:00:00 GMT"
+  (java-time.zone/zoned-date-time 1970 "UTC"))
+
+(defn- with-auth-tkt [response auth-tkt-value domain expiry-timestamp is-deleted]
+  (let [timestamp-string (java-time.format/format :rfc-1123-date-time (if is-deleted (unix-epoch) expiry-timestamp))]
```
esuomi, line 37: `unix-epoch` is a value now so don't call it.

esuomi (approved):
Overall I think this is a good minimum what can be done reasonably and cleanly. There's a few things that could be improved upon in the future:

 1. Use some actual cookie generation abstraction (I think we talked about those?) to get rid of string concatenation
 1. Middleware which resets cookies without Expire, this is mainly to onboard existing users to non-infinite cookie land
 1. Middleware which removes cookies from the request-response cycl if they're outdated

With that said, we can add those as subtasks and leave them to the next person in line to implement.

ote/src/clj/ote/services/login.clj
antti-nuortimo-solita: Fixed :)

ote/src/clj/ote/services/login.clj
antti-nuortimo-solita: Fixed :)

esuomi:
Closing this as time moved past it and too much has changed since.

PR #1387, opened 2022-01-28
FINAP-64/unroll parts of the query to adjust how prices are fetched so that the front page always populates as many entries as possible

esuomi:
That's a mouthful for a header...anyway, it says what's up.


antti-nuortimo-solita (approved):
Looks good to me :ship:

PR #1444, opened 2022-05-06
create new materialized view for looking up location relations easily

esuomi:
this is going to be use for drastic improvement of taxi statistics' performance


database/src/main/resources/db/migration/V1_200__materialized_view_for_location_lookups.sql
```diff
@@ -0,0 +1,51 @@
...
+       c.namefin    AS child_namefin,
+       c.nameswe    AS child_nameswe
+  FROM places p
+           JOIN places AS c ON (ST_Within(c.location, p.location) = TRUE OR
+                                ST_Overlaps(c.location, p.location) = TRUE)
+ WHERE p.type = 'finnish-municipality'
+   AND c.type = 'finnish-postal'
+WITH DATA;
+
+CREATE INDEX location_relations_p_type ON location_relations (parent_type);
```
rckopteri, line 39: Are the coming select queries only using one parameter at the time? I'm wondering if these indexes should be concatenated together like: 
CREATE INDEX location_relations_ptype_pnamefin ON location_relations (parent_type, parent_namefin);

esuomi: Usage is going to be `parent_type + parent_namefin + child_namefin` so yeah, I'll add one more index for this specific triplet 👍

PR #1544, opened 2025-01-10
Finap 260 postgres update

rckopteri:
Updating all relevant files to support postgresql 16.3 version.

Database update is simple:
- Drop detected-route-change_transit-change-date_fkey and gtfs_package_transport-service-id_fkey foreinkeys.
- Update 141 migration file
- Update vhecksum to schema_version table 
- Update database via AWS Console.
Those changes are hard to find from this pr.

Allmos all change changes are related to local docker update.


esuomi (approved):
Looks alright to me 👍 I'm still a bit perplexed by why the migration stopped working, but if it works with this, then great.

PR #1551, opened 2025-02-07
Use newest gtfs package in service-viewer instead of last successfull…

rckopteri:
…y converted package.

This changes situation for those interfaces that are using gtfs package which is broken. If the gtfs package is ok then user won't see any changes. If the package is broken we will now display the correct status.


esuomi (approved):
A bit surprisingly low amount of code for this change 👍

PR #1557, opened 2025-02-21
Finap 264 add netex interace support to vaco intg

rckopteri:


ote/src/cljs/ote/views/service_viewer.cljs
```diff
@@ -338,7 +338,8 @@
                     [:div
                      ; conversion status badge, falls back to entry badge if converter is unknown
                      [:img {:src    (str (:api-base-url tis-vaco) "/api/badge/" public-id (condp = (:name converter)
-                                                                                            "gtfs2netex.fintraffic" "/netex.entur"))
+                                                                                            "gtfs2netex.fintraffic" "/netex.entur"
+                                                                                            "netex2gtfs.entur" "/netex.entur"))
```
esuomi, line 342: This should probably be
```suggestion
                                                                                            "netex2gtfs.entur" "/gtfs.canonical"))
```
as the interesting validated conversion result is for the GTFS, not NeTEx. This has an impact on the related label as well though, instead of _"NeTEx-muunnoksen tila"_ it should probably be either a bit more abstract such as _"Muunnoksen tila"_, or somehow conversion target format aware.

PR #1591, opened 2026-05-08
Create a simple job in jenkins

solita-juhohaa:
This is created so that we can try out the Jenkins job deployment


solita-juhohaa:
This was unnecessary

